PR #36, opened 2022-08-04
feat: added optimal combination hierarchical reconciliation method

dluuo:
Added optimal combination hierarchical reconciliation method
[Link to paper here](https://robjhyndman.com/publications/hierarchical/)
Method currently supports 
- ordinary least squares (weighting matrix is simply identity matrix)
- structural weighted least squares(weighting matrix created from S matrix)
- can also input own weighting matrix for least squares


FedericoGarza:
Hey @dluuo!

Thanks a lot for the PR completing the hierarchical forecast catalog. We found that `OptimalCombination` from 2011 has an equivalence to the ols/wls MinTrace variants.

Since some people have already asked for this method, I think keeping the `OptimalComb` class with pointers to corresponding MinTrace methods for completeness will be fantastic. Could you make the change? :)

Your finding on the pseudo inverse for `wls_struct` will also be an excellent way to robustify `MinT`.

hierarchicalforecast/methods.py
```diff
@@ -324,6 +324,42 @@ def reconcile(
...
+                         method=method)
+
+# Cell
+class OptimalCombination:
+
+    def __init__(
+            self,
+            method: str # Least Squares Weighting Method
+        ):
+        self.method = method
```
FedericoGarza, line 346: It would be nice to have an assert here to consider just the optimal combination methods (`ols` and `wls_struct`).

hierarchicalforecast/methods.py
```diff
@@ -324,6 +324,42 @@ def reconcile(
...
+                         y_insample=y_insample,
+                         y_hat_insample=y_hat_insample,
+                         method=method)
+
+# Cell
+class OptimalCombination:
+
+    def __init__(
+            self,
+            method: str # Least Squares Weighting Method
```
FedericoGarza, line 344: Could you please add the allowed methods in the description? (after the `#` symbol).

FedericoGarza (changes requested):
Nice work, @dluuo! I left some comments

FedericoGarza:
I also noticed that the documentation for the new class is missing. You can generate it using `nbdev_build_docs`. The file `docs/methods.html` should be modified after executing that command. 

Finally, could you please clean the outputs of the notebook? You can do this by selecting `Edit -> Clear All Ouputs`.

hierarchicalforecast/methods.py
dluuo: fixed!

hierarchicalforecast/methods.py
dluuo: added!

FedericoGarza:
Thank you @dluuo! Welcome to hierarchicalforecast :)

PR #78, opened 2022-10-19
[FEAT] Non-negative reconciliation

FedericoGarza:
This PR adds a non-negative reconciliation option to the MinTrace algorithm.
Also, an example showing how to use the new feature is included.
The nonnegative option gets better performance on the Wiki2 dataset using the 'ols' method of MinTrace
compared with its unconstrained version.

- feat: add first version nonnegative rec
- feat: add nonnegative docs
- fix: build fn name to inclue nonnegative
- feat: add quadprog dep
- fix: use quadprog lib to avoid sqrt cals
- feat: add add nonnegative tests integrations
- fix: nonngeative reconcilaition
- feat: add nonnegative example


hierarchicalforecast/methods.py
```diff
@@ -461,8 +462,33 @@ def min_trace(S: np.ndarray,
     if any(eigenvalues < 1e-8):
         raise Exception(f'min_trace ({method}) needs covariance matrix to be positive definite.')
     
-    R = S.T @ np.linalg.pinv(W)
-    P = np.linalg.pinv(R @ S) @ R
+    W_inv = np.linalg.pinv(W)
+    if nonnegative:
+        if bootstrap:
+            raise Exception('nonnegative reconciliation is not compaitble with bootstrap forecasts')
```
kdgutier, line 468: compaitble

hierarchicalforecast/methods.py
```diff
@@ -461,8 +462,33 @@ def min_trace(S: np.ndarray,
...
+                                          axis=0, 
+                                          arr=y_hat)
+        y_hat = S @ bottom_fcts
+        return bottom_up(S=S, y_hat=y_hat, 
+                         idx_bottom=idx_bottom, 
+                         sigmah=sigmah, level=level)
+        #return {'mean': S @ bottom_fcts}
+    else:
+        # compute P for free reconciliation
+        R = S.T @ np.linalg.pinv(W)
```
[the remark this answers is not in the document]
kdgutier: Are these changes implementing the 
the [Optimal non-negative forecast reconciliation paper](https://robjhyndman.com/publications/nnmint/)?

If so, can we add the references to the docstrings?

hierarchicalforecast/methods.py
```diff
@@ -508,6 +541,7 @@ def reconcile(self,
         `y_hat`: Forecast values of size (`base`, `horizon`).<br>
         `y_insample`: Insample values of size (`base`, `insample_size`). Only used by `wls_var`, `mint_cov`, `mint_shrink`<br>
         `y_hat_insample`: Insample fitted values of size (`base`, `insample_size`). Only used by `wls_var`, `mint_cov`, `mint_shrink`<br>
+        `idx_bottom`: Indices corresponding to the bottom level of `S`, size (`bottom`).<br>
```
kdgutier, line 544: Would it be possible to infer the idx_bottom indexes from the S_[a+b, b] matrix?
something like  `idx_bottom = n.arange(a+b)[a:]`

If we commit to ordered numpy arrays, things might simplify.

Additionally the PERMBU method is operating under the assumption of hierarchically ordered numpy arrays.

hierarchicalforecast/methods.py
```diff
@@ -419,6 +418,8 @@ def min_trace(S: np.ndarray,
               y_insample: np.ndarray,
               y_hat_insample: np.ndarray,
               method: str,
+              idx_bottom: List[int] = None,
```
kdgutier, line 421: Would it be possible to infer the idx_bottom indexes from the S_[a+b, b] matrix?
something like  `idx_bottom = n.arange(a+b)[a:]`

If we commit to ordered numpy arrays, things might simplify.

Additionally the PERMBU method is operating under the assumption of hierarchically ordered numpy arrays.

hierarchicalforecast/methods.py
```diff
@@ -541,7 +579,9 @@ def optimal_combination(S: np.ndarray,
     return min_trace(S=S, y_hat=y_hat, 
                      y_insample=y_insample,
                      y_hat_insample=y_hat_insample,
-                     method=method, sigmah=sigmah, level=level,
+                     method=method, idx_bottom=idx_bottom,
```
kdgutier, line 582: Same comment on the need for `idx_bottom`

settings.ini
```diff
@@ -15,7 +15,7 @@ language = English
 custom_sidebar = True
 license = apache2
 status = 2
-requirements = numpy numba pandas scikit-learn statsmodels mycolorpy
+requirements = numpy numba pandas scikit-learn statsmodels mycolorpy quadprog
```
kdgutier, line 18: I suggest to stay only with scipy's nnls

hierarchicalforecast/methods.py
```diff
@@ -461,8 +462,33 @@ def min_trace(S: np.ndarray,
...
-    R = S.T @ np.linalg.pinv(W)
-    P = np.linalg.pinv(R @ S) @ R
+    W_inv = np.linalg.pinv(W)
+    if nonnegative:
+        if bootstrap:
+            raise Exception('nonnegative reconciliation is not compaitble with bootstrap forecasts')
+        if idx_bottom is None:
+            raise Exception('idx_bottom needed for nonnegative reconciliation')
+        # compute P for nonnegative reconciliation
+        warnings.warn('Replacing negative forecasts with zero.')
```
kdgutier, line 472: I suppose we are using a simplified version of NN reconciliation, deviating from its [original R implementation](https://github.com/earowang/hts/blob/master/R/bpv.R).

If that is the case can we comment the line of thought and differences versus R's version in this section?

hierarchicalforecast/methods.py
```diff
@@ -481,22 +507,29 @@ class MinTrace:
     
     **Parameters:**<br>
     `method`: str, one of `ols`, `wls_struct`, `wls_var`, `mint_shrink`, `mint_cov`.<br>
+    `nonnegative`: bool, reconciled forecasts should be nonnegative?<br>
 
     **References:**<br>
     - [Wickramasuriya, S. L., Athanasopoulos, G., & Hyndman, R. J. (2019). \"Optimal forecast reconciliation for
     hierarchical and grouped time series through trace minimization\". Journal of the American Statistical Association, 
     114 , 804–819. doi:10.1080/01621459.2018.1448825.](https://robjhyndman.com/publications/mint/).
+    - [Wickramasuriya, S.L., Turlach, B.A. & Hyndman, R.J. (2020). \"Optimal non-negative
```
kdgutier, line 516: Is this character `\"` needed? I think simple quotation marks suffice.

FedericoGarza: I don't know. I just followed your convention. See the previous reference. Would you like to change each reference?

kdgutier: At some point I had some issues with Microsoft parsing the curly quotation marks, I might have overcompensated.
I am unsure now if simple quotations compile correctly in nbdev. Would be good to check again.

hierarchicalforecast/methods.py
```diff
@@ -11,10 +11,9 @@
 
 import numpy as np
 from numba import njit
-
+from quadprog import solve_qp
```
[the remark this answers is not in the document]
FedericoGarza: That's the first thing I tried. The problem is that we are dealing with weighted NNLS instead of pure NNLS. The weights are provided by the inverse of the variance-covariance matrix. As stated in the documentation, `scipy.optimize.nnls` solves the problem,

```
min x || Ax - b|| st x>=0
```

If we want to include weights, we have to solve the problem,

```
min x || W^(-1/2)Ax - W^(-1/2)b|| st x>=0
```

Observe that, in this case, we need to calculate the squared root of `W^(-1/2)`. When `W` is diagonal, for example, `ols` or `wls_struct`, there is no problem. But with other methods, such as `mint_shrink`, it's impossible to guarantee the positiveness of each entry in `W`.  

That's the reason behind the election of `quadprog`.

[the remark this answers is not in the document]
kdgutier: Would it be possible to use Cholesky matrix and premultiply NNL inputs?

hierarchicalforecast/methods.py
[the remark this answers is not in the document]
FedericoGarza: That's the plan. I'll be opening a new PR just for that improvement to follow the convention we made about modular changes.

hierarchicalforecast/methods.py
FedericoGarza: It's possible, but it's just safer not to infer them. Also, it's not obvious how to test the assumption of hierarchically ordered numpy arrays. The complexity inside the methods simplifies the user experience in the `HierarchicalReconciliation` class.

hierarchicalforecast/methods.py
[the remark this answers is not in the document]
FedericoGarza: Positive definite W =/> `W_{ij} > 0` for all i,j

hierarchicalforecast/methods.py
kdgutier: I think we could test the equality of the following arrays: `y_[a] = A y_[b]`. To test for hierarchically ordered arrays.
Assuming testable properties in the numpy arrays would allow us to simplify the code and make it more efficient.

hierarchicalforecast/methods.py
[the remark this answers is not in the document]
kdgutier: I am getting confused, are we actually computing the sqrt of the entries of the matrix?
I thought the algorithm was using the sqrt of the inverse matrix, ie. Cholesky's matrix
https://en.wikipedia.org/wiki/Square_root_of_a_matrix

hierarchicalforecast/methods.py
FedericoGarza: > I think we could test the equality of the following arrays: y_[a] = A y_[b] for hierarchically order property of the input arrays.

And then you will include a new argument, `y_[b]`?

FedericoGarza: I'm not changing the `idx_bottom` since it is used in other methods. If you want to test your idea feel free to open an issue and make a PR. The `idx_bottom` argument is compatible with the previous structure.

hierarchicalforecast/methods.py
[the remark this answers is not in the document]
FedericoGarza: hey @kdgutier. I've updated the code to include comments about the decision behind `quadprog`,

https://github.com/Nixtla/hierarchicalforecast/blob/a5c1a7968442984477aaaf764a592a840619d9ee/hierarchicalforecast/methods.py#L474-L489

Issue #92, opened 2022-10-24
[BUG] Min Trace `method="mint_shrink"`


FedericoGarza:
Users have reported issues trying to run `MinTrace(method=='mint_shrink')`.

![](https://uploads.linear.app/e4b3e9f8-7a1a-4606-9de5-3b814aaf1193/e219e1cc-1140-43a3-998a-9c581291ddd7/644e9ae9-9fe6-428f-82b1-63463ba3086d)

I suspect that it is related to datasets with time series of distinct lengths (in our examples all time series have the same length and the method works fine). In particular, I think that line `441` below is not working correctly. It should ignore the NAS in the residuals matrix generated by the difference in lengths, but perhaps it is not ignoring them.

[https://github.com/Nixtla/hierarchicalforecast/blob/2df1dd86da606f7cfeddd4d5f2b4b620a8eaed72/hierarchicalforecast/methods.py#L438-L457](https://github.com/Nixtla/hierarchicalforecast/blob/2df1dd86da606f7cfeddd4d5f2b4b620a8eaed72/hierarchicalforecast/methods.py#L438-L457)

> From [Linear-GitHub Sync](https://synclinear.com)

FedericoGarza:
According to the [docs](https://numpy.org/devdocs/reference/generated/numpy.ma.cov.html), it is the most likely reason since `allow_masked=True` simply propagates the mask but does not remove the nas. 

<img width="809" alt="image" src="https://user-images.githubusercontent.com/10517170/197646610-c9bc0ec2-d992-4ae2-8571-8fe9569a1484.png">

PR #97, opened 2022-10-28
MinTrace's protection to Schafer-Strimmer covariance, eliminated stat…

kdgutier:
Solved eigenvalue indetermination in the presence of small percentage of zero predictions.
- Added protections for single series with constant values / std=0.
- Added protections to standardize residuals when std=0.
- Final epsilon diagonal for improved numerical stability.
- Zero residuals in `MinTrace` warning.
- Added stress test unit test for percentage of zero residuals scenarios in core.ipynb
- Added single insample value MinTrace-shr covariance stress test in methods.ipynb (ridge diagonal solved this problem).

I took out the `statsmodels`, the only function we used from it was `cov2corr` that is particularly simple and does not merit such a big dependency.


hierarchicalforecast/core.py
```diff
@@ -92,6 +92,8 @@ def reconcile(self,
         uids = Y_hat_df.index.unique()
         # check if Y_hat_df has the same uids as S
         if len(S.index.difference(uids)) > 0 or len(Y_hat_df.index.difference(S.index.unique())) > 0:
+            print(len(S.index.difference(uids)))
+            print(len(Y_hat_df.index.difference(S.index.unique())))
```
FedericoGarza, line 96: Please add a description of what is being printed. Those prints should be included in the `Exception` statement. For multiple lines, the following syntax can be used,

```
raise Exception(
        'line 1 '
        'line 2 '
        'line 3 ' 
)
````

File: hierarchicalforecast/methods.py [hunk 2]
```diff
@@ -336,6 +335,25 @@ def crossprod(x):
...
+    """
+    cov = np.asanyarray(cov)
+    std_ = np.sqrt(np.diag(cov))
+    corr = cov / np.outer(std_, std_)
+    if return_std:
+        return corr, std_
+    else:
+        return corr
+
+# %% ../nbs/methods.ipynb 34
```
FedericoGarza, line 356: Cool. Is it necessary to include the `return_std` argument? I did not see it being used in any function.

hierarchicalforecast/methods.py
```diff
@@ -398,27 +416,53 @@ def reconcile(self,
...
+                # this version follows https://robjhyndman.com/papers/MinT.pdf
                 tar = np.diag(np.diag(covm))
+
+                # Protection: constant's correlation set to 0
                 corm = cov2corr(covm)
-                xs = np.divide(residuals, np.sqrt(np.diag(covm)))
+                corm = np.nan_to_num(corm, nan=0.0)
+
+                # Protection: standardized residuals 0 where residual_std=0
+                residual_std =  np.sqrt(np.diag(covm))
```
FedericoGarza, line 449: If you want to preserve the `return_std` argument of `cov2corr`, please let's use its output here instead of computing `residual_std` twice. Something like,

```
corm, residual_std = cov2corr(cov, return_std=True)
```

environment.yml
```diff
@@ -7,7 +7,6 @@ dependencies:
   - numba
   - pandas
   - scikit-learn
-  - statsmodels
```
FedericoGarza, line 10: Please let's remove `statsmodels` also from `settings.ini`:

https://github.com/Nixtla/hierarchicalforecast/blob/4a8426210d295cd52713aa54030055be7114e67c/settings.ini#L18

The `environment.yml` file works for development, but `settings.ini` contains the actual requirements once the library is released.

FedericoGarza: Also, `statsmodels` is being installed/required here: https://github.com/Nixtla/hierarchicalforecast/blob/main/nbs/examples/AustralianDomesticTourism.ipynb.

Could you please remove it from there as well?

FedericoGarza (approved):
I have left some comments. As we are not testing the `nbs/examples` files automatically, please just check that we are able to recover the previous outputs/replications. I'll work on an automatic test to help future development.

hierarchicalforecast/core.py
FedericoGarza: Also, maybe we can avoid doing the calculation of the differences twice. Something like,

```
S_diff_uids  = S.index.difference(uids)
Y_hat_df_diff_uids = Y_hat_df.index.difference(S.index.unique())

if len(S_diff_uids) > 0 or len(Y_hat_df_diff_uids) > 0:
    raise Exception(print lens of S and Y_hat diffs)
```

FedericoGarza (commented):
Also, if this PR is going close https://github.com/Nixtla/hierarchicalforecast/issues/92, please let's add a test where it is verified that before the changes, the same error arises, and now with these changes, it is solved. 

See, for example, [this](https://github.com/Nixtla/hierarchicalforecast/pull/63/files#diff-f2047fa0bb3b650057cce0ade160f07da6e68afcee78c24ebc6c4d66a7450afe) PR. 

A user had an error using `HierarchicalEvaluation` when `h=1`, then a test was included showing that it works. In particular, these lines,

image

This particular case is a little bit tricky because we don't have the actual data. I'd start trying to reproduce the error with a small example. Just to be sure that these changes are actually solving the problem.

FedericoGarza:
> Also, if this PR is going close #92, please let's add a test where it is verified that before the changes, the same error arises, and now with these changes, it is solved.
> 
> See, for example, [this](https://github.com/Nixtla/hierarchicalforecast/pull/63/files#diff-f2047fa0bb3b650057cce0ade160f07da6e68afcee78c24ebc6c4d66a7450afe) PR.
> 
> A user had an error using `HierarchicalEvaluation` when `h=1`, then a test was included showing that it works. In particular, these lines,
> 
> image
> 
> This particular case is a little bit tricky because we don't have the actual data. I'd start trying to reproduce the error with a small example. Just to be sure that these changes are actually solving the problem.

Just some thoughts about the test:

An element of the diagonal of the variance-covariance matrix would be zero if a time series has only one value, so to reproduce the error, we could use the objects used for testing purposes in `nbs/methods.ipynb` as follows,

```
diff_len_y_insample = S @ y_bottom
diff_len_y_hat_insample = S @ y_hat_bottom_insample
diff_len_y_insample[-1, :-1] = np.nan
diff_len_y_hat_insample[-1, :-1] = np.nan
print(diff_len_y_insample)
cls_min_trace = MinTrace(method='mint_shrink')
cls_min_trace(
    S=S, 
    y_hat=S @ y_hat_bottom, 
    y_insample=diff_len_y_insample,
    y_hat_insample=diff_len_y_hat_insample,
    idx_bottom=idx_bottom
)
```

The output of the previous code is,

image

This small example recovers the #92 bug. Observe that if we use `ols`, there is no problem (as expected),

image

The user's data has time series with a single value.

image

hierarchicalforecast/core.py
kdgutier: Thanks I added the recommendations and complimented with extra insample `Y_df` protection.

hierarchicalforecast/methods.py
kdgutier: Nice catch, used it now.

environment.yml
kdgutier: Updated the `settings.ini`

hierarchicalforecast/methods.py
kdgutier: ibid.

PR #107, opened 2022-10-29
[FIX] Docs source links

### LINKED ISSUE #72: Source code links in documentation not working
When I open the documentation and click on the source code links, it leads to a 404 webpage error on github.

FedericoGarza:
Fixes #72﻿


mergenthaler (approved):
Great! I also noticed that the "Report an Issue" functionality on the documentation is not working. I think it links to a page that does not exist due to the templates that we have.

FedericoGarza:
> Great! I also noticed that the "Report an Issue" functionality on the documentation is not working. I think it links to a page that does not exist due to the templates that we have.

Nice catch! It looks like it was a general problem for all the links in the documentation because the base url was wrong. I tested it locally, and the fix solved that problem as well.🙌

PR #108, opened 2022-10-30
faster utils' `aggregation` 

kdgutier:
Improved `aggregation` function speed (now 20x faster), added unit tests to utils.ipynb
- Minor fix on `plot_hierarchical_predictions_gap` summation matrix attribute.
- Changed implicit pandas index sort operations for numpy aggregation.
- Restricted categorial encoder summation matrix responsibilities to upper level.
- Added unit tests for strictly hierarchical structure.
- To smooth transition and benchmarking, kept `aggregation_before` and `aggregation`.

PR #109, opened 2022-11-02
Fast TopDown/BottomUp

kdgutier:
- Added option of sparse reconciliation to `BottomUp`, and `TopDown` methods.
- Approach resulted in 15x speed gains.
- Added unit test comparing equality versus previous version.

Pending:
The `ERM` method by design creates a sparse P matrix, we should be able to see the same speed gains during inference.  Check if gains come from clever ordering of matrix operations.
If so update all such matrix operations accordingly.


FedericoGarza (changes requested):
Please keep the changes as modular and clean as possible. Some changes from [this](https://github.com/Nixtla/hierarchicalforecast/pull/108) PR are included in this new PR. Please separate them.

nbs/methods.ipynb
```diff
@@ -44,6 +44,7 @@
     "\n",
```
FedericoGarza, line 44: <pre class="ql-syntax" spellcheck="false">Line #17.            res = {'mean': np.matmul(S @ P, y_hat)}
</pre><p><code>np.matmul(S @ P, y_hat)</code>  is equivalent to <code>S.dot(P.dot(y_hat))</code>  when using numpy arrays? If that's the case, we could use <code>{'mean': S.dot(P.dot(y_hat))}</code> for all cases and leave the transformation to sparse arrays to the methods (to avoid adding a new argument).</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Nixtla/hierarchicalforecast/pull/109/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='5'/>

FedericoGarza, line 44: <pre class="ql-syntax" spellcheck="false">Line #24.                      sparsity: Optional[bool] = True):
</pre><p><code>sparsity</code> could be an attribute of the class instead of an extra argument. If <code>self.sparsity=True</code> , then <code>P</code> and <code>S</code> could be sent to <code>_reconcile</code> as sparse matrices.      </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Nixtla/hierarchicalforecast/pull/109/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='7'/>

FedericoGarza, line 44: <p>The <code>nbs/methods.ipynb</code> file is intended to work only with numpy arrays. If you want to test a large summing matrix, please create it from scratch using numpy instead of downloading and importing a dataframe. Since the inclusion of <code>CodeTimer</code> is not reviewed yet, you could test the performance of the change using the <code>time</code> module:</p><p> </p><pre class="ql-syntax" spellcheck="false">from time import time

# performace wo sparsity
init_wo_sparsity = time()
cls_bu = BottomUp()
cls_bu.reconcile(...)
end_wo_sparsity = time()

# performance sparsity
init_sparsity = time()
cls_bu = BottomUp(sparsity=True)
cls_bu.reconcile(...)
end_sparsity = time()

# test sparsity helps
assert (end_sparsity - init_sparsity) &lt; (end_wo_sparsity - init_wo_sparsity)
</pre>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Nixtla/hierarchicalforecast/pull/109/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='11'/>

FedericoGarza, line 44: <p>See previous comment on the <code>sparsity</code> attribute. If it works, it should be included in all the methods.    </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Nixtla/hierarchicalforecast/pull/109/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='23'/>

FedericoGarza (changes requested):
I like the idea. I think it would be better if the possibility to use sparse matrices is an argument of `HierarchicalReconciliation` instead of an argument/attribute of each class. For example, `HierarchicalReconciliation(sparsity=True, ...)`. In that case, the transformation of `S` is paid only once and all methods receive the same transformed matrix. 
Subsequently, the `_reconcile` function could determine if the matrix S is sparse and in that case transform P.

nbs/methods.ipynb
kdgutier: Sounds good, I added an [issue](https://github.com/Nixtla/hierarchicalforecast/issues/110) to change the corresponding matrix multiplication for the dot method.

I would keep the tests and the sparsity parameter momentarily to track these and other method's speed gains. Currently is a default parameter that is not shown to the higher-level methods.

kdgutier: My thought on skipping the `self.sparsity` attribute is to avoid extra complexity to the methods.  Some reconciliation methods might report gains consistently while other don't. 

We might want to make the selection a default.

kdgutier:
While doing internal testing on computation speed I realized that speed gains come from two sources:
1. Ordering of matrix multiplications `(S  P) y_[a,b]` vs `S ( P y_[a,b])`
2. Sparsity of `S` and `P` (that comes with conversion fixed cost)

Creation of sparse `S` should be done in `HierarchicalForecast.reconcile` **once**.
While the creation of sparse `P` should be done **once** for methods `BottomUp`, `TopDown`, `ERM-lasso`.
Rather than in every call to the `methods._reconcile` function.

PR #114, opened 2022-11-24
FEAT: add installation guide

mergenthaler:


kdgutier (approved):
I am thinking to avoid the ```> Install ``` comment to avoid the curly letters.

PR #167, opened 2023-01-23
[FIX] Use micromamba instead of miniconda (CI)

FedericoGarza:
This PR fixes the CI errors on Ubuntu due to miniconda.﻿


kdgutier (approved):
Thanks

PR #170, opened 2023-01-31
Omit expensive linear algebra when not necessary in MinTrace

mcsqr:
Context: We're experimenting with the package for rather high number of time series. Above 10k series it becomes this becomes somewhat cumbersome. Removing the unnecessary linalg invocations for the diagonal-only methods in MinTrace makes the execution somewhat (3-4 times) faster.

Note 1: I also made a version which avoids instantiating the diagonal matrix W at all and only keeps the diagonal in memory, but that's somewhat cumbersome to align with your treatment of missing values/nan-s in the covariance matrix, so I'm not commiting it here. BTW, I think your treatment of missing values is somewhat flawed: in case there are missing values/nan-s in the time series your implementation uses masking and in this way it returns _some_ numbers but, at least in our situation, they're _really_ bad. I personally think it's better to just fail in such cases.

Note 2: once going to _really_ high number of time series (say >100k) the matrix inversion in MinTrace is really killing it in terms of computing time and memory requirements. However, since one only needs the application of this matrix on a vector, one could get away without the explicit inversion and just use an iterative solver like conjugate gradients or similar. I didn't try to implement it as yet, but if there's some interest from your side to make the method work for really big data, perhaps we could collaborate here.


kdgutier:
Hi @mcsqr,

Thanks for the comments, and your contribution.
Our current `MinTrace` computational efficiency has lots of room for improvement.

Do you want to continue this PR?
Here are the contribution instructions:
https://github.com/mcsqr/hierarchicalforecast/blob/main/CONTRIBUTING.md

mcsqr:
Hi Kin, in this PR there was a mixup of my github.com account and my company's github account (which prevented me from signing the CLA), I made another one with the fix. BTW I think I followed all the steps in the contributing.md

PR #171, opened 2023-01-31
Omit expensive linear algebra when not necessary in MinTrace

mcsqr:
Context: We're experimenting with the package for rather high number of time series. Above 10k series it becomes this becomes somewhat cumbersome. Removing the unnecessary linalg invocations for the diagonal-only methods in MinTrace makes the execution somewhat (3-4 times) faster.

Note 1: I also made a version which avoids instantiating the non-diagonal parts matrix W at all and only keeps the diagonal in memory, but that's somewhat cumbersome to align with your treatment of missing values/nan-s in the covariance matrix, so I'm not commiting it here. BTW, I think your treatment of missing values is somewhat flawed: in case there are missing values/nan-s in the time series your implementation uses masking and in this way it returns _some_ numbers but, at least in our situation, they're really bad. I personally think it's better to just fail in such cases.

Note 2: once going to _really_ high number of time series (say >100k) the matrix inversion in MinTrace is really killing it in terms of computing time and memory requirements. However, since one only needs the application of this matrix on a vector, one could get away without the explicit inversion and just use an iterative solver like conjugate gradients or similar. I didn't try to implement it as yet, but if there's some interest from your side to make the method work for really big data, perhaps we could collaborate here.


FedericoGarza (approved):
LGTM! Thank you for this PR, @mcsqr. 

We are very interested in to make the method work for big data. Please feel free to open a PR with the necessary improvements. We'll be happy to review them.

kdgutier:
Hey @mcsqr,

Would you be able to look at this PR: https://github.com/Nixtla/hierarchicalforecast/pull/172
with a new computation of MinTrace's `P` matrix?
In theory its computational cost should be orders of magnitude faster.

PR #172, opened 2023-02-03
[FEAT] Efficient MinTrace

kdgutier:
Added the efficient implementation of MinTrace projection matrix.
We efficiently compute reconciliation matrix with
Equation 10 from https://robjhyndman.com/papers/MinT.pdf
New Computational complexity O(Na^3) vs O((Na+Nb)^3)

This comes at the expense of the deprecation in MinTrace of the use of `idx_bottom`.
In principle that argument should not be needed conditional on the presorting already included within the `core.reconcile` method.


candalfigomoro:
Is there any particular reason why this PR was not merged? Does this change create any problems?

kdgutier:
> Is there any particular reason why this PR was not merged? Does this change create any problems?

We need to deprecate the use of `idx_bottom` parameter on the rest of the methods.

jmoralez:
Superseded by #264

PR #181, opened 2023-04-17
[FEAT] Core Numeric Type and Null Protections

dluuo:
When a Y_hat_df is passed into HierarchicalForecast.reconcile() without numeric types, it leads to a silent bug where the reconcile takes extremely long and then throws an error.
I added a type check in the .reconcile  (core.ipynb) which throws an exception if the Y_hat_df does not have numeric types
I also added a NaN/None check in the .reconcile (core.ipynb)  which throws an exception if the Y_hat_df has null values

I also added  a NaN/None check in the aggregate function (utils.ipynb) which throws an exception if the Y_hat_df has null values


kdgutier (commented):
Thanks for the contribution.
- Can you add a `test_fail` unit test simulating the expected behavior of the protection? For a reference check the example on "different series S and Y_hat_df" in the same `nbs/core.ipynb` notebook.
- Let's add a Null protection to the target variable too.

kdgutier (approved):

PR #188, opened 2023-04-24
[FEAT] NeuralForecast Compatibility and Example Notebook

dluuo:
The Neuralforecast library outputs dataframes with a "[Model]-median" column, so I added some logic to handle the "-median" column. 
I also added an additional example notebook which demonstrates how to use HF with NF and MLF.


hierarchicalforecast/core.py
```diff
@@ -45,9 +45,14 @@ def _reverse_engineer_sigmah(Y_hat_df, y_hat, model_name):
...
-    drop_cols = ['ds', 'y'] if 'y' in Y_hat_df.columns else ['ds']
+
+    drop_cols = ['ds']
+    if 'y' in Y_hat_df.columns:
+        drop_cols.append('y')
+    if model_name+'-median' in Y_hat_df.columns:
+        drop_cols.append(model_name+'-median')
     model_names = Y_hat_df.drop(columns=drop_cols, axis=1).columns.to_list()
-    pi_model_names = [name for name in model_names if ('-lo' in name or '-hi' in name)]
+    pi_model_names = [name for name in model_names if ('-lo' in name or '-hi' in name or '-median' in name)]
```
kdgutier, line 55: Is this '-median' needed conditional on dropping it from model_names?

kdgutier (changes requested):
Pretty cool work.
Is it possible to include the `MLFrameworksExample.ipynb` notebook in the `sidebar.yml` for it to appear in the documentation.

This change in the `sidebar.yml` should suffice.
```
- section: ML Forecast Reconciliation
  contents:
  - examples/MLFrameworksExample.ipynb
```

hierarchicalforecast/core.py
dluuo: You are right it is not needed, removed

kdgutier (approved):

PR #190, opened 2023-05-11
[FIX] Aggregate unbalanced datasets

### LINKED ISSUE #189: `aggregate` adds leading zeros to series with different dates
The `aggregate` function adds leading zeros to datasets with different dates per time series.  Here's a minimal example:
```
import pandas as pd
import statsforecast.models as sfm
import hierarchicalforecast.methods as hfm
from statsforecast.utils import generate_series
from statsforecast import StatsForecast
from hierarchicalforecast.utils import aggregate
from hierarchicalforecast.core import HierarchicalReconciliation

max_tenure = 24
dates = pd.date_range(start='2019-01-31', freq='M', periods=max_tenure)
cohort_tenure = [24, 23, 22, 21]

ts_list = []

# Create ts for each cohort
for i in range(len(cohort_tenure)):
    ts_list.append(
        generate_series(n_series=1, freq='M', min_length=cohort_tenure[i], max_length=cohort_tenure[i]).reset_index() \
            .assign(ult=i) \
            .assign(ds=dates[-cohort_tenure[i]:]) \
            .drop(columns=['unique_id'])
    )
df = pd.concat(ts_list, ignore_index=True)

# Create categories
df.loc[df['ult'] < 2, 'pen'] = 'a'
df.loc[df['ult'] >= 2, 'pen'] = 'b'
# Note that unique id requires strings
df['ult'] = df['ult'].astype(str)

hier_levels = [
    ['pen'],
    ['pen', 'ult'],
]

hier_df, S_df, tags = aggregate(df=df, spec=hier_levels)
hier_df = hier_df.reset_index()
    # .query("unique_id.str.split('/').str[0] <= ds.dt.strftime('%Y-%m')")
print('S_df.shape', S_df.shape)
print('hier_df.shape', hier_df.shape)
```

If you query the 3rd cohort, we should see dates starting with 2019-03-31
```
df.query("ult == '2'")
```

But if you query `hier_df`, the output of aggregate, you'll see dates starting from 2019-01-31, the earliest date in the dataset.
```
hier_df.query("unique_id.str.split('/').str[-1] == '2'")
```

If you remove the leading zero's, `reconcile` fails because `forecast_fitted_values` cannot be reshaped into length of `S_df'.

FedericoGarza:
This PR fixes #189. 
Tests were added comparing the fixes with the deprecated function `aggregate_before` (which works as expected in this case).﻿


NudnikShpilkis:
@FedericoGarza looks like the change to `aggregate` works, but `reconcile` still fails.  Following my example from earlier
```
train_df = hier_df.query("ds <= @pd.to_datetime('2019-12-31')")
test_df = hier_df.query("ds > @pd.to_datetime('2019-12-31')")

fcst = StatsForecast(
    models=[
        sfm.AutoARIMA(season_length=12, alias='ARIMA'),
    ],
    freq='M',
    n_jobs=-1,
)

hrec = HierarchicalReconciliation(
    reconcilers=[
        hfm.BottomUp(),
        hfm.MinTrace(method='mint_shrink'),
    ]
)

fcst.fit(df=train_df)
fcst_df = fcst.forecast(h=12, fitted=True)
fitted_df = fcst.forecast_fitted_values()

fcst_df = hrec.reconcile(
    Y_hat_df=fcst_df,
    Y_df=fitted_df,
    S=S_df,
    tags=tags,
)
```

We get `ValueError: cannot reshape array of size 64 into shape (6,newaxis)` because`fitted_df` have a length of 64, while `S_df` has a length of 6.  You'll need to rewrite `reconcile` to address the reshaping.

NudnikShpilkis:
Not a bug per se, but I'd clarify the exception for not including `Y_df` in `reconcile`.  The type hints imply it's always optional, but it's only optional for bootstrap or permbu methods.  So the exception "you need to pass `Y_df`" is confusing in light of the type hint.

candalfigomoro:
My current workaround is to create a copy of train_df without the leading zeros and use that for the forecasting model. For reconcile() I still use the version with the leading zeros.

Note that aggregate() also creates all-zero time series for missing hierarchical combinations, so you also have to add all-zero forecasts to Y_hat_df/fcst_df for unique_ids artificially created by aggregate().

FedericoGarza:
hey @NudnikShpilkis and @candalfigomoro! Thank you for your feedback! I've updated the branch to consider this case in the `reconcile` method. Could you try uninstalling and installing the library again from this branch? :)

Thank you!

NudnikShpilkis:
@FedericoGarza looks like the code now runs without error.  I haven't tested for validity of predictions though.

cchallu (approved):

PR #195, opened 2023-06-02
[FEAT] Probabilistic Forecasting Util Functions

dluuo:
I added three functions to utils.ipynb that are using for probabilistic forecasting.
The main one is samples_to_quantiles_df(..), which given a numpy array of forecasts of samples, and quantiles/levels, it finds the quantiles of the sample forecasts and creates a dataframe where each columns are the forecasts corresponding to a quantile.

I also added level_to_outputs and quantile_to_outputs, which are helper functions imported from neuralforecast that given a level or quantiles list, outputs a array of names corresponding to those levels/quantiles.


hierarchicalforecast/utils.py
```diff
@@ -68,7 +68,75 @@ def cov2corr(cov, return_std=False):
     else:
         return corr
 
+# %% ../nbs/utils.ipynb 7
+# convert levels to output quantile names
+def level_to_outputs(level):
```
kdgutier, line 73: Let's add the following docstring:
"""
Converts list of levels into output names matching StatsForecast and NeuralForecast methods.

**Parameters:**<br>
`level`: int list [0,100]. Probability levels for prediction intervals.

**Returns:**<br>
`output_names`: str list. String list with output column names.
"""

hierarchicalforecast/utils.py
```diff
@@ -68,7 +68,75 @@ def cov2corr(cov, return_std=False):
...
+    # Add default median
+    quantiles = np.concatenate([np.array([50]), quantiles])
+    quantiles = torch.Tensor(quantiles) / 100
+    output_names = list(np.array(output_names)[sort_idx])
+    output_names.insert(0, '-median')
+    
+    return quantiles, output_names
+
+# convert quantiles to output quantile names
+def quantiles_to_outputs(quantiles):
```
kdgutier, line 89: Let's add the following docstring:

"""
Converts list of quantiles into output names matching StatsForecast and NeuralForecast methods.

**Parameters:**<br>
`quantiles`: float list [0., 1.]. Alternative to level, quantiles to estimate from y distribution.

**Returns:**<br>
`output_names`: str list. String list with output column names.
"""

hierarchicalforecast/utils.py
```diff
@@ -68,7 +68,75 @@ def cov2corr(cov, return_std=False):
...
+        elif q>.50:
+            output_names.append(f'-hi-{np.round(100-200*(1-q),2)}')
+        else:
+            output_names.append('-median')
+    return quantiles, output_names
+
 # %% ../nbs/utils.ipynb 8
+# given input array of sample forecasts and inptut quantiles/levels, 
+# output a Pandas Dataframe with columns of quantile predictions
+def _to_quantiles_df(samples,
```
kdgutier, line 103: Here is a draft for the docstring of this function, let's polish it:
""" Transform Samples into HierarchicalForecast input.

Auxiliary function to create compatible HierarchicalForecast input `Y_hat_df` dataframe.

**Parameters:**<br>
`samples`: numpy array. Samples from forecast distribution of shape [n_series, n_samples, horizon].<br>
`unique_ids`: .<br>
`dates`: .<br>
`quantiles`: .<br>
`level`: .<br>
`model_name`: .<br>

**Returns:**<br>
`Y_hat_df`: pd.DataFrame, With base quantile forecasts with columns `ds` and models to reconcile indexed by `unique_id`.<br>
"""

hierarchicalforecast/utils.py
```diff
@@ -68,7 +68,75 @@ def cov2corr(cov, return_std=False):
...
+# output a Pandas Dataframe with columns of quantile predictions
+def _to_quantiles_df(samples, 
+                     unique_ids, 
+                     dates, 
+                     quantiles = None,
+                     level = None, 
+                     model_name = "model"):
+    
+    # Get the shape of the array
+    N, S, H = samples.shape
```
kdgutier, line 111: I recommend to use `n_series, n_samples, horizon` constants, because HierarchicalForecast already uses `S` and `H` for other constants. It may be confusing.

kdgutier (changes requested):

hierarchicalforecast/utils.py
```diff
@@ -68,7 +68,106 @@ def cov2corr(cov, return_std=False):
...
+    """
+    qs = sum([[50-l/2, 50+l/2] for l in level], [])
+    output_names = sum([[f'-lo-{l}', f'-hi-{l}'] for l in level], [])
+
+    sort_idx = np.argsort(qs)
+    quantiles = np.array(qs)[sort_idx]
+
+    # Add default median
+    quantiles = np.concatenate([np.array([50]), quantiles])
+    quantiles = torch.Tensor(quantiles) / 100
```
FedericoGarza, line 90: Currently, the library does not need `torch` as a dependency. Is there a particular reason to return and use `quantiles` as a torch tensor? Using the function could throw errors since `torch` is not being imported in this module. Changing these lines to 

```
quantiles = np.concatenate([np.array([50]), quantiles]) / 100
```

should work.

FedericoGarza (changes requested):
hey @dluuo! Nice work so far! I just left a comment related to the usage of `torch` inside the `level_to_outputs` function. 

Also, it might be a good moment to include some unit tests to catch potential errors in the new functions. You can see [here](https://github.com/Nixtla/statsforecast/blob/main/nbs/core.ipynb) some examples. 

The main idea is to think about the expected output of the function for a specific case, write that expected output, and then compare the actual and expected outputs. For example, in the case of `level_to_outputs` a unit test could be something like:

```python
test_eq(
     level_to_outputs([80, 90]),
     ([0.05, 0.1, 0.5, 0.9, 0.95], ['lo-90', 'lo-80', 'median', 'hi-80', 'hi-90'])
)
```

hierarchicalforecast/utils.py
dluuo: fixed, thanks!

FedericoGarza (approved):

PR #203, opened 2023-06-09
[Doc] Added baseline experiments and minor protection to Normality reconciler

kdgutier:
- Added baseline experiments.
- Minor catch to warnings in Normality reconciler to avoid redundant 'RuntimeWarning: covariance is not positive-semidefinite.' message. By definition, the sampled multivariate Normal is rank-defficient.


experiments/hierarchical_baselines/src/run_baselines.py
```diff
@@ -0,0 +1,318 @@
...
+
+    LEVEL = np.arange(0, 100, 2)
+    qs = [[50-lv/2, 50+lv/2] for lv in LEVEL]
+    QUANTILES = np.sort(np.concatenate(qs)/100)
+
+    # Run experiments
+    crps_results_list = []
+    msse_results_list = []
+    #for dataset in ['Labour', 'Traffic', 'TourismSmall', 'TourismLarge', 'Wiki2']:
+    for dataset in ['OldTraffic', 'OldTourismLarge']:
```
FedericoGarza, line 296: Is there a particular reason to exclude the remaining datasets? It might be better to have an argument called `--dataset` that can receive any available dataset.

experiments/hierarchical_baselines/src/run_baselines.py
```diff
@@ -0,0 +1,318 @@
...
+    print(f'\n {intervals_method.upper()} statistical baselines evaluation \n')
+
+    LEVEL = np.arange(0, 100, 2)
+    qs = [[50-lv/2, 50+lv/2] for lv in LEVEL]
+    QUANTILES = np.sort(np.concatenate(qs)/100)
+
+    # Run experiments
+    crps_results_list = []
+    msse_results_list = []
+    #for dataset in ['Labour', 'Traffic', 'TourismSmall', 'TourismLarge', 'Wiki2']:
```
FedericoGarza, line 295: What's the difference between `OldTraffic` and `Traffic`, `OldTourismLarge` and `TourismLarge`? Is there documentation about the differences?

hierarchicalforecast/probabilistic_methods.py
```diff
@@ -79,8 +80,11 @@ def get_samples(self, num_samples: int=None):
         n_series, n_horizon = self.y_hat.shape
         samples = np.empty(shape=(num_samples, n_series, n_horizon))
         for t in range(n_horizon):
-            partial_samples = state.multivariate_normal(mean=self.SP @ self.y_hat[:,t],
-                                                  cov=self.cov_rec[t], size=num_samples)
+            with warnings.catch_warnings():
+                # Avoid 'RuntimeWarning: covariance is not positive-semidefinite.'
+                # By definition the multivariate distribution is not full-rank
+                partial_samples = state.multivariate_normal(mean=self.SP @ self.y_hat[:,t],
+                                                    cov=self.cov_rec[t], size=num_samples)
```
FedericoGarza, line 87: This is nice and really helpful in avoiding unwanted prints.🚀

FedericoGarza (changes requested):
Nice @kdgutier! The `catch_warnings` it’s an excellent addition.  🙌

I would suggest making these kinds of changes in separate PRs so we can be more modular and precise in the release notes of the library. For example, ideally, in the release notes, we will have a note on the `catch_warnings` and a different note about adding the baseline experiments. 

I just left a couple of questions and suggestions.

experiments/hierarchical_baselines/src/run_baselines.py
kdgutier: Thanks for the suggestion; I will remember it for the next PR.
This time I saw the opportunity to make a minor improvement while working on the baselines.

I added the `--dataset` script argument.

experiments/hierarchical_baselines/src/run_baselines.py
kdgutier: I added an explanation to the difference of datasets in the `DatasetsForecast` PR:
https://github.com/Nixtla/datasetsforecast/pull/22

FedericoGarza (approved):
Thank you @kdgutier! Nice improvements. The community will benefit a lot from the hierarchical baselines scripts.

PR #210, opened 2023-06-15
Sparse Reconciliation

mcsqr:
Using sparse matrices to reduce the computational time and memory appetite of a subset of the reconciliation methods. 

The PR contains MinTraceSparse which supports the diagonal methods (ols, wls_struct, and wls_var), the speed-up is very considerable for datasets of ~2k-20k time series (the most I checked so far).

It also contains BottomUpSparse, which doesn't really show speed-up until 20k, but should hopefully help with even bigger datasets.

Note: This version still doesn't guarantee that no O(N^2) dense matrix is instantiated, it could probably be checked/improved by testing on even bigger datasets.

Also likely some docs and tests are missing, tbd.


kdgutier:
Thanks, @mcsqr 
This is awesome; we will take a look.

From the execution of the CI tests, we will need to fix the scikit-learn version to have the updated sparse OneHotEncoder. This would need to be done here:
https://github.com/Nixtla/hierarchicalforecast/blob/main/environment.yml
https://github.com/Nixtla/hierarchicalforecast/blob/main/settings.ini
I don't know if we will need to increase min_python > 3.7

mcsqr:
CI tests are fixed. Note that sparse in OneHotEncoder is deprecated and scheduled for removal, so at some point it'll also break, but for now it looks ok.

hierarchicalforecast/core.py
```diff
@@ -249,6 +249,19 @@ def reconcile(self,
         self.sample_names = {}
         for reconcile_fn, name_copy in zip(self.reconcilers, self.orig_reconcilers):
             reconcile_fn_name = _build_fn_name(name_copy)
+
+            # A pretty crude way to recognize a sparse method:
```
kdgutier, line 253: Since sparsifying is computationally demanding, performing the sparsification of the S matrix outside of the for loop is convenient.

What is your opinion on declaring once `S_sparse` and sharing it across BottomUpSparse and MinTraceSparse-variants?

Pseudo-code would be:
```
any_sparse = "Sparse" in any method's name.

if any_sparse:
	S_sparse = declare_S_sparse
.
.
.
for method in methods:
        reconcilers_args['S'] = S
	if method_is_sparse:
		reconcilers_args['S'] = S_sparse
        
        everything else is the same
```

To keep the code clean, we can declare a function to determine whether particular methods are sparse and use it in the declaration of `S_sparse` and within the for loop.

kdgutier (commented):
Awesome @mcsqr,

My most important comment is on the declaration of S_sparse in core.py.
Thanks a lot for your contribution.

hierarchicalforecast/core.py
mcsqr: Note: here I am _not_ sparsifying the data. This: `S_df.sparse.to_coo()` will throw an exception if S_df is not already a sparse DataFrame, and then we fall back to getting the dense version, and only sparsify S inside the MinTraceSparse/BottomUpSparse method.

hierarchicalforecast/utils.py
```diff
@@ -258,27 +260,40 @@ def _to_summing_dataframe(df: pd.DataFrame,
...
-    # TODO: option to only operate with sparse matrices
     bottom_ids = list(S_df.unique_id)
     del S_df['unique_id']
     categories = [S_df[col].unique() for col in S_df.columns]
     tags = dict(zip(S_df.columns, categories))
     tags[bottom_col] = bottom_ids
 
-    encoder = OneHotEncoder(categories=categories,
-                            sparse=False, dtype=np.float32)
+    encoder = OneHotEncoder(
```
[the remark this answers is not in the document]
mcsqr: The only change in OneHotEncoder is that they changed the argument name from `sparse` to `sparse_output` (likely due to the conflict with scipy.sparse package; I actually got a crazy nasty error because of this name conflict once). I am avoiding creating S as dense matrix as much as possible, as for really large datasets (which I'm ultimately aiming at) it will oom immediately.

hierarchicalforecast/core.py
kdgutier: I see; then, let's keep it as it is. But I would rather have it fail than overwrite it with the dense fallback.
This way commit to execute Sparse methods as intended by the user.

Can we change the except to:
```
raise Exception(f'Sparse methods require a sparse summation dataframe `S_df`, \
, see [aggregate method](https://nixtla.github.io/hierarchicalforecast/utils.html#aggregate)')
```

mcsqr: This will still work fine, it will sparsify it inside the method, e.g. here: hierarchicalforecast/methods.py:838 (this is still fast, I can provide timing in my 20k dataset later). How about a warning? If something really bad is going to happen (e.g. oom), then it will happen already at the level of creating dense S in the `aggregate`

kdgutier: If we allow for a dense fallback, then the remark on the unnecessary computation of sparsification becomes true.

If sparsification is made at hierarchicalforecast/methods.py:838, it would be executed every time we call a MinTraceSparse-variant.

mcsqr: But yeah, I guess we could do it before the loop. I still think it's premature optimisation, but whatever, I'll code it and see.

kdgutier: Regarding computational gains on MinTrace, I have this PR that you might find useful too:
https://github.com/Nixtla/hierarchicalforecast/pull/172

kdgutier: Regarding how you timed the methods, I would decompose it into Sparsification fixed cost + MinTrace cost.
I have found in the past that once sparsification is covered. You really see a benefit.

mcsqr: I pushed a change.

For the timing: since we are discussing single seconds here, I'm not sure if you're aware about how much improvement this PR brings on large datasets. I never actually managed to time the baseline MinTrace on my full 23k dataset, as it takes so long, but on half of it (11k time series), these are the timings for reconciliation: 
MinTraceSparse: 2.74s
Standard MinTrace: 442.73s
On larger datasets, the improvement is expected to be even more substantial, due to better computational complexity.
I talked with Fede about trying to time M5 dataset reconciliation, I'll try to do it later this week.

kdgutier (approved):

hierarchicalforecast/core.py
kdgutier: Awesome @mcsqr 

Thanks a lot for this contribution.

Issue #211, opened 2023-06-15
Broken aggregate function in main branch


candalfigomoro:
### What happened + What you expected to happen

If I use hierarchicalforecast v0.3.0 there's a problem with "unbalanced" time series, see: https://github.com/Nixtla/hierarchicalforecast/issues/189

However, if you exclude this specific issue, it works fine.

If I use the latest version from the main branch (maybe after this commit? https://github.com/Nixtla/hierarchicalforecast/commit/c1072171e074470e97d343e3fbb0188f68509f5d) I have huge issues.

I have unbalanced time series.

If I set the new "is_balanced" parameter to False (the default behavior), the returned dataframe has many missing dates (ds). For some unique_ids I expect many years of data but I just get 2 rows.

If I set the new "is_balanced" parameter to True, I get a reshaping error.

The same code works with version 0.3.0. I had to revert to the released 0.3.0 version because the version in the main branch is not usable for me.

### Versions / Dependencies

hierarchicalforecast main/master branch
Python 3.10
Linux
pandas 2.0.2

### Reproduction script

`Y_df, S_df, tags = aggregate(Y_df, spec)`

Unfortunately, I can't share my data set.

### Issue Severity

None

candalfigomoro:
Just to add some context:

I have time series with the same ending date but different starting dates.

I think it's ok for reconciliation to fill missing dates with zeros.

However, for model training data I don't want to add leading zeros to my time series.

NudnikShpilkis:
Doesn't `aggregate` have a `is_balanced` parameter not a `is_unbalanced`? The function assumes the data is _not_ balanced by default.

candalfigomoro:
> Doesn't `aggregate` have a `is_balanced` parameter not a `is_unbalanced`? The function assumes the data is _not_ balanced by default.

Yes I misspelled in the message above. The parameter is called is_balanced and the default value is False. I edited the message above. The issue is the same.

patleg78:
I was looking into the source code for new version and old version
here I find some the order of code logic
new version with sparse (say snippet1 )
"    Y_bottom_df = Y_bottom_df.groupby(['unique_id', 'ds'])['y'].sum().reset_index()
    Y_bottom_df.unique_id = Y_bottom_df.unique_id.astype('category')
    Y_bottom_df.unique_id = Y_bottom_df.unique_id.cat.set_categories(S_df.columns)
"

here we are aggregating first and then getting column names as index for Y_bottom_df.

[...]

PR #229, opened 2023-08-23
Fixes for large datasets

mcsqr:
Some minor but somewhat helpful changes:
* This avoids invoking `S.values` which creates the dense matrix from S when `sparse_s = True` in the `aggregate` function. This avoids OOM errors for very large datasets (I've used it [the `aggregate` function] successfully on a 400k time-series dataset after the change).
* It also removes a warning coming from using the deprecated `sparse` argument. Backwards compatibility with old sklearn libraries is maintained.

**Excluded from this PR:**
* It gives a warning whenever Y_df has less time series than S_df at the end of the `aggregate` function, as current behaviour is quite unintuitive and leads to users' confusion.
[EDIT: Aaah, ok, it seems that this behaviour was introduced as a result of this PR here: Nixtla/hierarchicalforecast/pull/190, see also this issue: Nixtla/hierarchicalforecast/issues/211 . What was the reason again to move this line `Y_bottom_df = Y_bottom_df.groupby(['unique_id', 'ds'])['y'].sum().reset_index()` two lines up?]
[EDIT 2: Perhaps there should be an additional argument with an option whether to prepend the shorter time series with zeros, or not?]

### DIFF
--- a/nbs/utils.ipynb
+++ b/nbs/utils.ipynb
@@ -580,6 +580,7 @@
   {
    "cell_type": "code",
    "execution_count": null,
+   "metadata": {},
    "outputs": [],
    "source": [
     "#| hide\n",
@@ -613,10 +614,7 @@
     "\n",
     "test_eq(Y_df.columns, Y_df_sparse.columns)\n",
     "test_eq(Y_df.index, Y_df_sparse.index)"
-   ],
-   "metadata": {
-    "collapsed": false
-   }
+   ]
   },
   {
    "cell_type": "markdown",


jmoralez:
Hey. Sorry for taking so long to review, I'll take a look now.

The errors weren't exactly related to python 3.7 but rather scikit-learn >= 1.2, [which was released in Dec 2022](https://pypi.org/project/scikit-learn/1.2.0/), not so long ago to stop supporting it. Can you please use something like the following instead?
```python
try:
    encoder = OneHotEncoder(categories=categories, sparse_output=sparse_s, dtype=np.float32)
except TypeError:  # sklearn < 1.2
    encoder = OneHotEncoder(categories=categories, sparse=sparse_s, dtype=np.float32)
```

mcsqr:
Hi Jose, yeah, thanks for the tip. I did it.

Note: the currently biggest point of discussion is the last bullet point in the PR description. It seems that the change you (Fede) have introduced in PR #190 is very unintuitive to users. I'm not sure what's the right way to handle this, but you have some issues open and I also had complaints from my collaborators. The warning I added is just a small plaster on a big wound. But this could also be taken out of this PR if desired.

hierarchicalforecast/utils.py
```diff
@@ -70,22 +70,29 @@ def cov2corr(cov, return_std=False):
...
-    encoder = OneHotEncoder(categories=categories, sparse=False, dtype=np.float32)
+
+    try:
+        encoder = OneHotEncoder(categories=categories, sparse_output=sparse_s, dtype=np.float32)
+    except TypeError:  # sklearn < 1.2
+        encoder = OneHotEncoder(categories=categories, sparse=sparse_s, dtype=np.float32)
+
     S = encoder.fit_transform(S_df).T
+    # TODO: This likely still instantiates a dense matrix:
     S = pd.DataFrame(S, index=chain(*categories), columns=cats_bottom)
```
jmoralez, line 87: I think we should use the [recommended constructor](https://pandas.pydata.org/docs/user_guide/sparse.html#interaction-with-scipy-sparse), something like:
```python
if sparse_s:
    df_constructor = pd.DataFrame.sparse.from_spmatrix
else:
    df_constructor = pd.DataFrame
S = df_constructor(S, index=chain(*categories), columns=cats_bottom)
```

hierarchicalforecast/utils.py
```diff
@@ -271,7 +285,16 @@ def aggregate(
...
                 y = np.concatenate([y_agg, y_bottom], axis=0)))
-    Y_df = Y_df.set_index('unique_id').dropna()
+
+    Y_df = Y_df.set_index("unique_id").dropna()
+
+    if S_df.index.nunique() != Y_df.index.nunique():
+        print(
+            "Warning: Some time series were dropped from Y_df. "
+            "Are you sure your time series don't have missing entries?"
+        )
```
jmoralez, line 295: I think we can leave this for a follow up PR where we try to directly address #211

hierarchicalforecast/utils.py
```diff
@@ -255,9 +266,12 @@ def aggregate(
 
     #------------------------------- Aggregation -------------------------------#
     n_agg = S_df.shape[0] - S_df.shape[1]
-    Agg = S_df.values[:n_agg, :]
-    y_bottom = balanced_df.y.values
+    if sparse_s:
+        Agg = S_df.sparse.to_coo().tocsr()[:n_agg, :]
```
jmoralez, line 270: Can you add a test in the notebook to verify that this is working as expected? Could be just a small example with sparse_s set to both True and False and verify that the outputs are the same.

jmoralez (changes requested):
Just some minor comments. Also can you please revert the changes to python 3.7? The library is still compatible with it and we want to support it for a little big longer.

hierarchicalforecast/utils.py
mcsqr: Yeah, I can remove it from this PR. Note however that it's a major annoyance for the users.

hierarchicalforecast/utils.py
mcsqr: Done

hierarchicalforecast/utils.py
mcsqr: Done

hierarchicalforecast/utils.py
mcsqr: Done

mcsqr:
@jmoralez I think all done.

jmoralez (approved):
Thanks a lot! LGTM

PR #242, opened 2023-10-17
keep only observed combinations in aggregate

jmoralez:
Adds `observed=True` to the groupby that aggregates the series in the `aggregate` function so that the output has only the observed combinations in case the columns are categoricals.


AzulGarza (approved):
Nice @jmoralez!

PR #249, opened 2023-11-10
support parallel MinTrace optimization

jmoralez:
Since the `solve_qp` function [uses the nogil context manager](https://github.com/quadprog/quadprog/blob/a7f4fa1a59bb962ebd24b8d3ac84fc53c51409f5/quadprog/quadprog.pyx#L95-L100) it's possible to parallelize the optimization problems using a thread pool, which will solve the optimization problems for each horizon in parallel.

This adds the `num_threads` argument to `MinTrace`, which if set to a number greater than 1 will spawn a thread pool and submit the work there.

Also removes deprecation warnings about the sparse argument in the OneHotEncoder.


AzulGarza (changes requested):
Awesome @jmoralez! Could we test that the same results can be recovered using `num_threads=1` and `num_threads>1`? We could add the test in a cell with `#| eval: false`, just to test it while developing.

AzulGarza (approved):

PR #263, opened 2024-02-29
Update README.md

B8ni:
mis-type


elephaint:
Thanks! I addressed this in #296

PR #280, opened 2024-08-16
[FEAT] Efficient Schafer-Strimmer for MinT

elephaint:
This PR implements an optimized version of the Schafer-Strimmer shrunk empirical covariance algorithm for `MinT`.

### 1. Up to 60x faster 
First, the result is that we can perform `MinT` reconciliation much faster, as demonstrated in the table below, where `mint_legacy` is the old version and `mint` is the new version, with the number in brackets denoting whether the forecasts contain NaNs (True or False) and the number of bottom-level timeseries reconciled (i.e., `20`, `200`, `2000`). An improvement of up to 60x (on my machine, i7-12700k) is seen (no NaNs, 2000 timeseries). 

Note that the NaN-version is about 3x slower than the version that doesn't have to deal with NaNs, which is due to all the masking involved when dealing with NaNs. However, it is still significantly faster than the legacy version (that doesn't handle NaNs properly).

Screenshot 2024-09-02 113916

### 2. Forecasting performance
Forecasting performance is similar:

Screenshot 2024-09-02 113002

### 3. Improved NaN handling
This PR improves NaN-handling (thanks to @christophertitchen for the suggestions!), leading to more cases that MinT can handle. For example, introducing 20% NaN values in the Australian Tourism forecasts couldn't be handled by the old version but the new version will give results nearly identical to the no-nan case: 

Screenshot 2024-09-02 113218

Code for introducing NaNs in the Australian Tourism example:

```
# Randomly set NaNs
nan_fraction = 0.2
Y_fitted_df = Y_fitted_df.reset_index()
Y_fitted_df_sample_idx = Y_fitted_df.sample(frac=nan_fraction).index
Y_fitted_df.loc[Y_fitted_df_sample_idx, ["y", "ETS"]] = np.nan
Y_fitted_df = Y_fitted_df.set_index("unique_id")
```


hierarchicalforecast/methods.py
```diff
@@ -895,6 +913,59 @@ def fit_predict(self,
...
+            sum_sq_emp_corr += np.square(factor_emp_corr * w_mean)
+
+    # Calculate shrinkage intensity 
+    shrinkage = max(min(sum_var_emp_corr / (sum_sq_emp_corr + epsilon), 1.0), 0.0)
+    # Calculate shrunk covariance estimate
+    emp_cov_diag = np.diag(emp_cov)
+    W = (1.0 - shrinkage) * emp_cov + mint_shr_ridge
+
+    # Fill diagonal with original empirical covariance diagonal
+    np.fill_diagonal(W, emp_cov_diag)
```
christophertitchen, line 965: Ah, this is a beautiful and creative approach to complete the convex transformation by taking advantage of the shrinkage target being a diagonal matrix of sample variances. :heart_eyes:

hierarchicalforecast/methods.py
```diff
@@ -895,6 +913,59 @@ def fit_predict(self,
...
+            Xs_j_mean = np.mean(Xs_j)
+            # Empirical covariance
+            emp_cov[i, j] = factor_emp_corr * np.mean(X_i * X_j)
+            # Sum off-diagonal variance of empirical correlation
+            w = (Xs_i - Xs_i_mean) * (Xs_j - Xs_j_mean)
+            w_mean = np.mean(w)
+            sum_var_emp_corr += (i != j) * factor_var_emp_cor * np.sum(np.square(w - w_mean))
+            # Sum squared empirical correlation (off-diagonal correction made by initializing 
+            # with -n_timeseries, so (i != j) not necessary here)
+            sum_sq_emp_corr += np.square(factor_emp_corr * w_mean)
```
christophertitchen, line 956:
```suggestion
        for j in range(i + 1):
            # Calculate standardized residuals
            X_j = residuals[j] - residuals_mean[j]
            Xs_j = X_j / (residuals_std[j] + epsilon)
            Xs_j_mean = np.mean(Xs_j)
            # Empirical covariance
            emp_cov[i, j] = emp_cov[j, i] = factor_emp_corr * np.mean(X_i * X_j)
            # Sum off-diagonal variance of empirical correlation
            w = (Xs_i - Xs_i_mean) * (Xs_j - Xs_j_mean)
            w_mean = np.mean(w)
            sum_var_emp_corr += 2 * (i != j) * factor_var_emp_cor * np.sum(np.square(w - w_mean))
            # Sum squared empirical correlation (off-diagonal correction made by initializing 
            # with -n_timeseries, so (i != j) not necessary here)
            sum_sq_emp_corr += 2 * np.square(factor_emp_corr * w_mean)
```

You can exploit the symmetry of the covariance matrix to reduce the number of iterations by $`\frac{1}{2}n\left( n + 1 \right)`$ in the serialised region of the parallel loop, as $`\text{cov}\left( y_{i}, \, y_{j} \right) \equiv \text{cov}\left( y_{j}, \, y_{i} \right)`$.

elephaint: Again a great idea, thanks!

hierarchicalforecast/methods.py
elephaint: Thanks!

hierarchicalforecast/methods.py
```diff
@@ -743,16 +747,30 @@ def _get_PW_matrices(self,
 
                 # Protection: final ridge diagonal protection
                 W = (lmd * tar + (1 - lmd) * covm) + self.mint_shr_ridge
-
                 UtW = Ut @ W
+            elif self.method == 'mint_shrink':
+                residuals_mean = np.nanmean(residuals, axis=0, dtype=np.float64)
+                residuals_std = np.maximum(np.nanstd(residuals, axis=0, dtype=np.float64), 1e-6)
+                safe_residuals = np.nan_to_num(residuals.T, copy=False)
```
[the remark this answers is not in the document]
elephaint: Curious to redo the benchmark when Numpy makes the release, nice work on that PR!

hierarchicalforecast/methods.py
elephaint: Numba doesn't complain, and I made a few other changes that improve the wall time, so again thanks!

elephaint:
@christophertitchen Again thanks for the remarks, I made a few other micro optimizations, in the end in total further reducing computation time by ~30%

christophertitchen (approved):
I am happy the comments were helpful!

The only further change I thought of would be to potentially modify the naïve approach when computing covariance and correlation to a pairwise approach for each time series, like what is done in ``stats::cov`` for R. This could better describe the relationship between pairs of time series, but honestly, it would slow down the generation of the shrunk $W$ and go against the spirit of what you were trying to achieve here by implementing a fast method, so I think it is perfectly fine as you have done it.

As a quick example to demonstrate, imagine monthly time series representing the units sold of products over the past six months. We take two of those products, $X$ and $Y$, with $Y$ being a NPI (new product introduction) which was implemented only four months prior. The residuals of a naïve forecast would be:

$`X = [(x_{t-6} - x_{t-7}), (x_{t-5} - x_{t-6}), (x_{t-4} - x_{t-5}), (x_{t-3} - x_{t-4}), (x_{t-2} - x_{t-3}), (x_{t-1} - x_{t-2})]`$

$`Y = [-, -, -, (y_{t-3} - y_{t-4}), (y_{t-2} - y_{t-3}), (y_{t-1} - y_{t-2})]`$

This approach would use an ``n_samples`` of $6$ and the the mean of $X$ would include all six observations. However, a pairwise approach would only use pairs of temporally aligned observations, so an ``n_samples`` of $3$, and the mean of $X$ would be $`\overline{X} = ((x_{t-3} - x_{t-4}) + (x_{t-2} - x_{t-3}) + (x_{t-1} - x_{t-2})) / 3`$, which would obviously lead to a different $`\text{cov}(X, Y)`$, $`\text{corr}(X, Y)`$, $`\sigma_{x}`$ for standardisation and so on.

```
not_mask = ~np.isnan(residuals)
```

```
...
for i in prange(n_timeseries):
    x = residuals[i]
    for j in range(i + 1):
        y = residuals[j]
        # Get the temporally aligned pairs of observations.
        mask = not_mask[i] & not_mask[j]
        n_samples = len(mask)
        # Check for insufficient observations for unbiased estimator.
        if n_samples - 1 == 0:
            ...
        x = x[mask]
        y = y[mask]
        # Calculate the pairwise sample means.
        mean_x = np.mean(x)
        mean_y = np.mean(y)
        ...
```

Anyway, I am not familiar with any literature on this topic or if there is any evidence to suggest it is significantly worse, and the legacy version also takes the same approach, so as I said, I favour the fast approach you have implemented here. Good job!

elephaint:
> I am happy the comments were helpful!
> 
> The only further change I thought of would be to potentially modify the naïve approach when computing covariance and correlation to a pairwise approach for each time series, like what is done in `stats::cov` for R. This could better describe the relationship between pairs of time series, but honestly, it would slow down the generation of the shrunk W and go against the spirit of what you were trying to achieve here by implementing a fast method, so I think it is perfectly fine as you have done it.
> 
> As a quick example to demonstrate, imagine monthly time series representing the units sold of products over the past six months. We take two of those products, X and Y , with Y being a NPI (new product introduction) which was implemented only four months prior. The residuals of a naïve forecast would be:
> 
> X = [ ( x t − 6 − x t − 7 ) , ( x t − 5 − x t − 6 ) , ( x t − 4 − x t − 5 ) , ( x t − 3 − x t − 4 ) , ( x t − 2 − x t − 3 ) , ( x t − 1 − x t − 2 ) ]
> 
> Y = [ − , − , − , ( y t − 3 − y t − 4 ) , ( y t − 2 − y t − 3 ) , ( y t − 1 − y t − 2 ) ]
> 
> This approach would use an `n_samples` of 6 and the the mean of X would include all six observations. However, a pairwise approach would only use pairs of temporally aligned observations, so an `n_samples` of 3 , and the mean of X would be X ― = ( ( x t − 3 − x t − 4 ) + ( x t − 2 − x t − 3 ) + ( x t − 1 − x t − 2 ) ) / 3 , which would obviously lead to a different cov ( X , Y ) , corr ( X , Y ) , σ x for standardisation and so on.
> 
> ```
> not_mask = ~np.isnan(residuals)
> ```
> 
> ```
> ...
> for i in prange(n_timeseries):
>     x = residuals[i]
>     for j in range(i + 1):
>         y = residuals[j]
>         # Get the temporally aligned pairs of observations.
>         mask = not_mask[i] & not_mask[j]
>         n_samples = len(mask)
>         # Check for insufficient observations for unbiased estimator.
>         if n_samples - 1 == 0:
>             ...
>         x = x[mask]
>         y = y[mask]
>         # Calculate the pairwise sample means.
>         mean_x = np.mean(x)
>         mean_y = np.mean(y)
>         ...
> ```
> 
> Anyway, I am not familiar with any literature on this topic or if there is any evidence to suggest it is significantly worse, and the legacy version also takes the same approach, so as I said, I favour the fast approach you have implemented here. Good job!

Didn't think about this, but it's a good point. I'll have a try and play around a bit with it :)

elephaint:
@christophertitchen Thanks for the suggestion for the temporal alignment - I included two versions, one that can handle NaNs in the way you describe. It's a bit slower but it enlarges the scope of MinT significantly, as it can handle much more cases without leading to ill-conditioned `W` matrices.

hierarchicalforecast/methods.py
```diff
@@ -988,6 +1005,116 @@ def fit_predict(self,
...
+    sum_sq_emp_corr = np.float64(0.0)
+    factor_emp_corr = np.float64(n_samples / (n_samples - 1))
+    factor_shrinkage = np.float64(1 / (n_samples * (n_samples - 1)))
+    epsilon = np.float64(2e-8)
+    for i in prange(n_timeseries):
+        # Mean of the standardized residuals
+        Xs_i_mean = np.mean(Xs[i])
+        for j in range(i + 1):
+            # Empirical covariance
+            W[i, j] = factor_emp_corr * np.mean(Xs[i] * Xs[j])
```
christophertitchen, line 1036: I think we may be inadvertently allocating $W$ as the correlation matrix here as this is the covariance of the standardised residuals, right? I thought ``Xs`` are the standardised deviations from the means. It might be worth checking if we are actually shrinking the correlation matrix instead of the covariance matrix by accident.

hierarchicalforecast/methods.py
```diff
@@ -822,7 +822,7 @@ def _get_PW_matrices(self,
                 corm = np.nan_to_num(corm, nan=0.0)
                 xs = np.divide(residuals, residual_std, 
                                out=np.zeros_like(residuals), where=residual_std!=0)
-
+                
                 xs = xs[~np.isnan(xs).any(axis=1), :]
                 v = (1 / (n * (n - 1))) * (crossprod(xs ** 2) - (1 / n) * (crossprod(xs) ** 2))
```
christophertitchen, line 827:
```suggestion
                v = (n / ((n - 1) ** 3)) * (crossprod(xs ** 2) - (1 / n) * (crossprod(xs) ** 2))
```

I know this is not changed in the PR, but is it me or does this look wrong? ``np.ma.cov`` gives an unbiased estimate by default, and this would only be valid for a biased estimate using $`N`$ degrees of freedom.

hierarchicalforecast/methods.py
```diff
@@ -837,6 +837,23 @@ def _get_PW_matrices(self,
...
+                UtW = Ut @ W
+            elif self.method == 'mint_shrink':
+                # Compute nans
+                nan_mask = np.isnan(residuals.T)
+                # Compute shrunk empirical covariance
+                if np.any(nan_mask):
+                    residuals_mean = np.nanmean(residuals, axis=0, dtype=np.float64, keepdims=True).T
+                    residuals_std = np.maximum(np.nanstd(residuals, axis=0, dtype=np.float64, keepdims=True), 1e-6).T
+                    safe_residuals = np.nan_to_num(residuals.T, copy=False)
+                    Xs = (safe_residuals - residuals_mean) / (residuals_std + 2e-8)
```
christophertitchen, line 849: The only issue I can see with this is that the samples of observations for each pair of time series may differ if the masks do not align, so the means would have to be calculated in the loop for each $i$ and $j$ before calculating the deviations and standardising. I know that unfortunately this would slow down the NaN version.

hierarchicalforecast/methods.py
```diff
@@ -988,6 +1005,116 @@ def fit_predict(self,
...
+    for i in prange(n_timeseries):
+        for j in range(i + 1):
+            if i != j:    
+                W[i, j] = W[j, i] = shrinkage * W[i, j] + mint_shr_ridge
+            else:
+                W[i, j] = W[j, i] = max(W[i, j], mint_shr_ridge)
+    return W
+
+@njit(parallel=True, fastmath=True, error_model="numpy")
+def _shrunk_covariance_schaferstrimmer_with_nans(Xs: np.ndarray, not_nan_mask: np.ndarray, mint_shr_ridge: float):
```
christophertitchen, line 1059: Thank you for implementing this version! :thumbsup:

christophertitchen (approved):
I added a few more thoughts, but I am honestly not too familiar with the Schäfer and Strimmer paper, so I will leave the rest in your more capable hands! :sweat_smile:

hierarchicalforecast/methods.py
elephaint: Yeah, I think I'm mixing it, I already had this feeling after one of the revisions I did. I need to see how to correct it, but good obvervation

hierarchicalforecast/methods.py
elephaint: I was under the impression that this is a consolidation of factors (see below), but I also remember that when I checked a while ago the unbiasedness isn't completely correct in the original implementation in R (our original follows the original Hyndman implementation from R). 

In the original Schafer-Strimmer algorithm:
```
factor_emp_corr = (n_samples / (n_samples - 1)
factor_var_emp_cor = (n_samples / (n_samples - 1)**3
for i ....
    for j ....
       sum_var_emp_corr += 2 * factor_var_emp_cor * np.sum(np.square(w - w_mean))
       sum_sq_emp_corr += 2 * np.square(factor_emp_corr * w_mean)

shrinkage = sum_var_emp_corr / (sum_sq_emp_corr + epsilon)
```

Now, we can equivalently write (we can ignore the epsilon, that's there for stability purposes):
```
factor_shrinkage = 1 / (n_samples * (n_samples - 1))
for i ....
    for j ....
       sum_var_emp_corr += np.sum(np.square(w - w_mean))
       sum_sq_emp_corr += w_mean**2

shrinkage = (factor_shrinkage * sum_var_emp_corr) / (sum_sq_emp_corr + epsilon)
```

I think I'll leave it as it is for now.

hierarchicalforecast/methods.py
elephaint: True, it needs to be calculated in-loop per pair, fixed that, thanks.

hierarchicalforecast/methods.py
elephaint: Solved it. This also explains the performance delta between the versions I was observing. Now the legacy version and new implementation give nearly identical results when tested on an actual dataset (synthetic was equal already).

elephaint:
@christophertitchen Again thanks for the thoughtful comments. 

I think I'll be deprecating the legacy version now that I get equivalent results across datasets.

christophertitchen:
> @christophertitchen Again thanks for the thoughtful comments.
> 
> I think I'll be deprecating the legacy version now that I get equivalent results across datasets.

@elephaint great job!

Also, regarding Hyndman's implementation, you were right.

https://github.com/earowang/hts/blob/1408ab2c5c40f1022e6957bcf8438aaefc8464bf/R/MinT.R#L24

```
  covm <- crossprod(x) / n
```

```
  v <- (1/(n * (n - 1))) * (crossprod(xs^2) - 1/n * (crossprod(xs))^2)
```

It uses the biased estimator, i.e. normalisation by $N$. We could have changed ``ddof`` in ``np.ma.cov`` to $0$ to copy this approach, but deprecating the legacy approach in favour of your superior one makes sense. 😉

settings.ini
```diff
@@ -15,7 +15,7 @@ language = English
 custom_sidebar = True
 license = apache2
 status = 2
-requirements = numpy numba pandas scikit-learn quadprog matplotlib tqdm
+requirements = numpy numba pandas scikit-learn quadprog matplotlib tqdm icc_rt;platform_system!="Darwin"
```
jmoralez, line 18: How/where is this used?

elephaint: This enables additional optimizations (SVML, short vector math library) for Numba for x86-64 systems, yielding significant speedup on those systems.

[See the Numba docs](https://numba.readthedocs.io/en/stable/user/performance-tips.html#intel-svml), although I see that as of recently they suggest to use a smaller package instead. I updated to use the smaller package (`intel-cmplr-lib-rt`) instead.

jmoralez: I see. I'm guessing we also need to restrict on the architecture, since this won't be available for ARM CPUs and I'm not sure if installing there will try to build from source. Can you please also add `and platform_machine=="x86_64"`? You can change the way these are read in `setup.py` to use commas for example to avoid having to collapse it.

setup.py
```diff
@@ -24,7 +24,7 @@
     '4 - Beta', '5 - Production/Stable', '6 - Mature', '7 - Inactive' ]
 py_versions = '3.8 3.9 3.10 3.11 3.12'.split()
 
-requirements = cfg.get('requirements','').split()
+requirements = cfg.get('requirements',',').split()
```
jmoralez:
```suggestion
requirements = cfg['requirements'].split(',')
```

elephaint: Ok, I'd already noticed the misstake with the comma, but I can remove the get too I guess

jmoralez (dismissed):

jmoralez (approved):

PR #294, opened 2024-10-09
ENH: Add support for exogenous variables in utils.aggregate

KuriaMaingi:
This change to the utility function will assist in instances where you need to generate your summation and Y_df but also want to retain any exogenous vars required for your forecast.

You will need to pass in a dictionary containing your exogenous vars and the Pandas agg functions you want applied against them.

I have currently hardcoded the list of acceptable agg_funcs but open to hear if there's a better way

### DIFF
--- a/hierarchicalforecast/utils.py
+++ b/hierarchicalforecast/utils.py
@@ -154,6 +154,7 @@ def _to_upper_hierarchy(bottom_split, bottom_values, upper_key):
 def aggregate(
     df: pd.DataFrame,
     spec: List[List[str]],
+    exog_vars: Dict[str, str] = None,
     is_balanced: bool = False,
     sparse_s: bool = False,
 ):
@@ -167,6 +168,9 @@ def aggregate(
         Dataframe with columns `['ds', 'y']` and columns to aggregate.
     spec : list of list of str
         List of levels. Each element of the list should contain a list of columns of `df` to aggregate.
+    exog_vars: dictionary of string keys & values
+        keys correspond to column names and the values represent the aggregation that will be applied to each column. Accepted values are those from Pandas aggregation Functions e.g. sum, np.mean etc.
+        acceptable_aggregations = {'sum', 'mean', 'median', 'min', 'max', 'count', 'std', 'var', 'first', 'last'}
     is_balanced : bool (default=False)
         Deprecated.
     sparse_s : bool (default=False)
@@ -190,6 +194,21 @@ def aggregate(
             "Don't set this argument to suppress this warning.",
             category=DeprecationWarning,
         )
+    
+    # Define acceptable aggregation functions
+    acceptable_aggregations = {
+        'sum', 'mean', 'median', 'min', 'max', 'count', 'std', 'var', 'first', 'last'
+    }    
+    # Check if exog_vars are present in df
+    if exog_vars is not None:
+        missing_vars = [var for var in exog_vars.keys() if var not in df.columns]
+        if missing_vars:
+            raise ValueError(f"The following exogenous variables are not present in the DataFrame: {', '.join(missing_vars)}")
+      
+        # Check if the aggregation functions are acceptable
+        invalid_aggregations = [agg for agg in exog_vars.values() if agg not in acceptable_aggregations]
+        if invalid_aggregations:
+            raise ValueError(f"The following aggregation functions are not acceptable: {', '.join(invalid_aggregations)}")       
             
     # compute aggregations and tags
     spec = sorted(spec, key=len)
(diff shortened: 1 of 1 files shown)


elephaint:
@KuriaMaingi Thanks for your work! I'll happily take a look :) 

We use nbdev, which means changes to the code should be made in source notebooks - in your case `nbs\utils.ipynb`. 

To set your environment best up to work on this, I'd advise to:

0. Clone the repository and go to the root directory in a Terminal window
1. Create a conda environment hierarchicalforecast: `conda create -n hierarchicalforecast python=3.10`
2. Activate the envionrment: `conda activate hierarchicalforecast`
3. Install the required packages: `conda env update -f environment.yml`
4. Install the locally cloned library editable: `pip install -e ".[dev]"`
5. Install git hooks: `nbdev_install_hooks`
6. Install pre-commit: `pre-commit install`

Now make your changes to the notebook, in your case `nbs\utils.ipynb`. Make sure to clean the notebook before exporting it (`Edit -> Clear All Outputs` in your IDE)

7. Build the library: `nbdev_export`
8. Use git add, commit and push commands to push the branch and create the PR onwards.

christophertitchen:
> @KuriaMaingi Thanks for your work! I'll happily take a look :)
> 
> We use nbdev, which means changes to the code should be made in source notebooks - in your case `nbs\utils.ipynb`.
> 
> To set your environment best up to work on this, I'd advise to:
> 
> 0. Clone the repository and go to the root directory in a Terminal window
> 1. Create a conda environment hierarchicalforecast: `conda create -n hierarchicalforecast python=3.10`
> 2. Activate the envionrment: `conda activate hierarchicalforecast`
> 3. Install the required packages: `conda env update -f environment.yml`
> 4. Install the locally cloned library editable: `pip install -e ".[dev]"`
> 5. Install git hooks: `nbdev_install_hooks`
> 6. Install pre-commit: `pre-commit install`
> 
> Now make your changes to the notebook, in your case `nbs\utils.ipynb`. Make sure to clean the notebook before exporting it (`Edit -> Clear All Outputs` in your IDE)
> 
> 7. Build the library: `nbdev_export`
> 8. Use git add, commit and push commands to push the branch and create the PR onwards.

@KuriaMaingi to add to the great summary above, you can also use the commands below before exporting.

- `nbdev_clean --clear_all` to double-check that all of the metadata and cell outputs are removed in the notebooks to avoid any merge conflicts.
- `nbdev_test --n_workers 1 --do_print --timing` to execute tests in the notebooks sequentially and report on the timings.

hierarchicalforecast/utils.py
```diff
@@ -190,14 +194,40 @@ def aggregate(
...
     aggs = []
     tags = {}
+    # Prepare the aggregation dictionary
+    agg_dict = {
+        "y": ("y", "sum")
+    }
+
+    # Add exog_vars to the aggregation dictionary if it is not None
+    if exog_vars is not None:
+        agg_dict.update({key: (key, exog_vars[key]) for key in exog_vars.keys()})
```
christophertitchen, line 225: Could you please give an example usage of `exog_vars` in this context?

I have not used pandas much lately, but I think that given your type signature of `Dict[str, str]`, you intend to have `exog_vars = {"col_a": "sum", "col_b": "sum"}`. However, this does not support multiple functions to aggregate a particular column as you are going down the named aggregation route.

A way around this will be either `exog_vars = {"col_a": ("sum", "mean")}` which will create a `MultiIndex`, or alternatively something like `exog_vars = {"col_a_sum": ("col_a", "sum"), "col_a_mean": ("col_a", "mean")}`. Either way, the distinction between the output column name for the aggregation and the column name to be aggregated will need to be made when inserting into `agg_dict` to avoid overwriting anything in this case.

hierarchicalforecast/utils.py
```diff
@@ -190,14 +194,40 @@ def aggregate(
             "Don't set this argument to suppress this warning.",
             category=DeprecationWarning,
         )
+    
+    # Define acceptable aggregation functions
+    acceptable_aggregations = {
+        'sum', 'mean', 'median', 'min', 'max', 'count', 'std', 'var', 'first', 'last'
+    }
```
christophertitchen, line 201: I do not think this is needed—we can just let pandas raise an `AttributeError` when aggregating rather than raising our own `ValueError`.

Plus, this gives us the flexibility to use custom (anonymous) functions rather than just string function names.

christophertitchen (commented):
Thank you for helping with the project, it is nice to have more contributors!

I left a few thoughts before you make the changes in the notebooks and export them.

KuriaMaingi:
Thanks all for the comments, I will close this and replace with a new PR following the preferred approach. Thanks

PR #295, opened 2024-10-10
[FEAT] Improve residuals-based reconciliation stability and faster ma.cov

elephaint:
- Adds faster ma.cov method to mint_cov
- Improves stability of residuals-based methods.
- Includes test for covariance matrices in MinT
- Includes benchmark that can be run for MinT and ma.cov, useful for future benchmarking


hierarchicalforecast/methods.py
```diff
@@ -966,6 +970,35 @@ def fit_predict(self,
...
     __call__ = fit_predict
 
+@njit(parallel=True, fastmath=True, error_model="numpy")
+def _ma_cov(residuals: np.ndarray, not_nan_mask: np.ndarray):
+    """Masked empirical covariance matrix.
+
+    :meta private:
+    """
+    n_timeseries = residuals.shape[0]
+    W = np.zeros((n_timeseries, n_timeseries), dtype=np.float64).T
```
christophertitchen, line 980:
```suggestion
    W = np.empty((n_timeseries, n_timeseries), dtype=np.float64)
```

It will be a tad faster to not initialise the values in the array as we are setting them all in the loop, plus no need for the transpose.

hierarchicalforecast/methods.py
```diff
@@ -966,6 +970,35 @@ def fit_predict(self,
...
+                # Masked residuals
+                residuals_i = residuals[i][not_nan_mask_ij]
+                residuals_j = residuals[j][not_nan_mask_ij]
+                residuals_i_mean = np.mean(residuals_i)
+                residuals_j_mean = np.mean(residuals_j)
+                X_i = (residuals_i - residuals_i_mean)
+                X_j = (residuals_j - residuals_j_mean)
+                # Empirical covariance
+                factor_emp_cov = np.float64(1 / (n_samples - 1))
+                W[i, j] = W[j, i] = factor_emp_cov * np.mean(X_i * X_j)
```
christophertitchen, line 998:
```suggestion
                W[i, j] = W[j, i] = factor_emp_cov * np.sum(X_i * X_j)
            else:
                W[i, j] = W[j, i] = np.nan
```

We need the sum of the product of the deviations as we are dividing by `n_samples - 1`, and `np.nan` for the cases without enough samples.

hierarchicalforecast/methods.py
```diff
@@ -814,22 +818,18 @@ def _get_PW_matrices(self,
...
-            except np.linalg.LinAlgError:
-                raise Exception(f'min_trace ({self.method}) needs covariance matrix to be positive definite. \n Please use another reconciliation method.')
-        else:
-            eigenvalues = np.diag(W)
-            if any(eigenvalues < 1e-8):
-                raise Exception(f'min_trace ({self.method}) needs covariance matrix to be positive definite.')
-            
-        P = (J - np.linalg.solve(UtW[:, n_aggs:] @ Ut.T[n_aggs:] + UtW[:, :n_aggs], UtW[:, n_aggs:] @ J.T[n_aggs:]).T @ Ut)
+        try:
+            P = (J - np.linalg.solve(UtW[:, n_aggs:] @ Ut.T[n_aggs:] + UtW[:, :n_aggs], UtW[:, n_aggs:] @ J.T[n_aggs:]).T @ Ut)
```
christophertitchen, line 827: I like this formulation for the non-diagonal $W$ methods!

christophertitchen (changes requested):
I left a few suggestions—good stuff!

hierarchicalforecast/methods.py
```diff
@@ -966,6 +970,35 @@ def fit_predict(self,
 
     __call__ = fit_predict
 
+@njit(parallel=True, fastmath=True, error_model="numpy")
```
christophertitchen, line 973:
```suggestion
@njit("f8[:, :](f8[:, :], b1[:, :])", nogil=True, cache=True, parallel=True, error_model="numpy", fastmath=True)
```

Do you think it would be worth it to use eager compilation by specifying the signature and caching? I guess in the vast majority of cases the function would only be called once to try out the `mint_cov` method, so most likely not.

Anyway, you may also want to experiment with releasing the GIL to see if it makes a difference when benchmarking as the function is parallelised.

christophertitchen (changes requested):

hierarchicalforecast/methods.py
elephaint: Thanks - did you test this? I wrote it like this because benchmarks showed the zero initialization + transpose is faster later on (it has to do with the memory lay out being more convenient for later use).

elephaint: Ran some benchmarks - only `nogil=True` seems to have an effect, so I'm making that change.

Base case:

image

Not transposing W on init:

image

Static typing + nogil + cache:

image

Nogil:

image

hierarchicalforecast/methods.py
elephaint: Thanks! I made the change for the mean, and added a test to make sure the empirical covariance is the same across methods (when calculating without NaNs)

`np.nan` for cases without enough samples is of course the technically correct method, but that will also give an error later on - I'd prefer to just set it to zero in those cases.

hierarchicalforecast/methods.py
elephaint: See earliier comment with benchmarks:
- Static typing generally slows it down, even when accounting for contiguity of arrays
- `nogil=True` makes sense
- `cache=True` makes sense

Thanks!

christophertitchen (approved):
I was so confused why you were transposing a square zero matrix but it makes complete sense now that you wanted to change the order! 😅

What are you using to benchmark there? It looks amazing and much nicer than `timeit`. Yes, if you want the output $W$ to have `0` instead of `np.nan`, your approach is definitely the best. I am surprised that the strict typing had any impact on the performance in this case, even when being really pedantic and specifying a row-major $W$ return with `f8[::1, :]`.

Anyway, the last thing I can think of would be to specify the Fortran order when allocating and initialising the array, but transpose is so fast as it only needs to swap the strides of the array, so it would probably only shave off a few nanoseconds or microseconds depending on the number of time series.

```suggestion
    W = np.zeros((n_timeseries, n_timeseries), dtype=np.float64, order="F")
```

LGTM! 🚀

hierarchicalforecast/methods.py
jmoralez: You can initialize it with fortran layout:
```python
np.empty((n_timeseries, n_timeseries), dtype=np.float64, order='F')
```

`empty` should be faster than `zeros`

elephaint:
> I was so confused why you were transposing a square zero matrix but it makes complete sense now that you wanted to change the order! 😅
> 
> What are you using to benchmark there? It looks amazing and much nicer than `timeit`. Yes, if you want the output W to have `0` instead of `np.nan`, your approach is definitely the best. I am surprised that the strict typing had any impact on the performance in this case, even when being really pedantic and specifying a row-major W return with `f8[::1, :]`.
> 
> Anyway, the last thing I can think of would be to specify the Fortran order when allocating and initialising the array, but transpose is so fast as it only needs to swap the strides of the array, so it would probably only shave off a few nanoseconds or microseconds depending on the number of time series.
> 
> ```
>     W = np.zeros((n_timeseries, n_timeseries), dtype=np.float64, order="F")
> ```
> 
> LGTM! 🚀

Thanks - I agree. The reason the `.T` is necessary is that Numba doesn't support the order-argument of Numpy functions (that's why I wrote it like that, it's a bit of a low-cost hack to get F-order in Numba).

marcopeix (approved):

PR #296, opened 2024-10-10
[FEAT] Add linting and pre-commit

elephaint:
- Adds linting to github actions
- Adds pre-commit to the repo
- Updates contributing guide to reflect our other repos
- Adds uv
- Fixes all the resulting mypy / ruff issues arising from applying these checks
- Bumps Python dependencies to 3.9, deprecating 3.8 support


.github/workflows/ci.yml
```diff
@@ -49,14 +27,14 @@ jobs:
...
-        uses: mamba-org/setup-micromamba@f8b8a1e23a26f60a44c853292711bacfd3eac822 # v1.9.0
+        uses: mamba-org/setup-micromamba@617811f69075e3fd3ae68ca64220ad065877f246 # v2.0.0
         with:
           environment-file: environment.yml
           create-args: python=${{ matrix.python-version }}
           cache-environment: true
 
       - name: Install pip requirements
-        run: pip install ".[dev]"
+        run: pip install uv && uv pip install --system ".[dev]"
```
christophertitchen, line 37: @elephaint you legend, this is the first time I have ever seen anybody use `uv` in the wild. Have you had much experience using it and would you recommend it? I want to give it a try over `poetry` and `pyenv`, but I want to avoid any potential headaches on work projects!

.github/workflows/ci.yml
```diff
@@ -49,14 +27,14 @@ jobs:
...
 
       - name: Run tests
-        run: nbdev_test
+        run: nbdev_test --do_print --timing --n_workers 0
```
christophertitchen, line 40: Oh, that is weird! I thought it would be `1` worker to run the unit tests sequentially with a single thread or process. Does `0` do the same thing here?

.github/workflows/lint.yml
```diff
@@ -0,0 +1,25 @@
...
+      - name: Set up python
+        uses: actions/setup-python@f677139bbe7f9c59b41e40162b753c062f5d49a3 # 5.2.0
+        with:
+          python-version: 3.8
+
+      - name: Install dependencies
+        run: pip install black nbdev==2.3.25 pre-commit
+
+      - name: Run pre-commit
+        run: pre-commit run --files hierarchicalforecast/*
```
christophertitchen, line 25: Oh dear, did we not have any pre-commit hooks? 🤯

.pre-commit-config.yaml
christophertitchen, line 1: LGTM! 👍

hierarchicalforecast/core.py
```diff
@@ -154,7 +155,7 @@ def _prepare_fit(self,
         model_names = [name for name in model_names if name not in pi_model_names]
         
         # TODO: Complete y_hat_insample protection
-        if intervals_method in ['bootstrap', 'permbu']:
+        if intervals_method in ['bootstrap', 'permbu'] and Y_df is not None:
```
christophertitchen, line 158: Good catch!

hierarchicalforecast/core.py
```diff
@@ -90,7 +91,7 @@ class HierarchicalReconciliation:
     [Rob J. Hyndman and George Athanasopoulos (2018). \"Forecasting principles and practice, Hierarchical and Grouped Series\".](https://otexts.com/fpp3/hierarchical.html)
     """
     def __init__(self,
-                 reconcilers: List[Callable]):
+                 reconcilers: List[HReconciler]):
```
christophertitchen, line 94: 👌

hierarchicalforecast/methods.py
christophertitchen, line 1: A solid clean-up job, good work!

christophertitchen (commented):
I only had time to go through some of it so far, sorry. It looks really good though!

.github/workflows/lint.yml
elephaint: Nope, I was giving the contributing instructions in the other PR and realized that we don't even support what I stated there, haha....

.github/workflows/ci.yml
elephaint: @jmoralez came with this; it's a rust-based package manager that seems mainly very much faster than pip / conda.

.github/workflows/ci.yml
elephaint: 0 means everything is run in the same (single) thread, see documentation of fastcore parallel

hierarchicalforecast/core.py
elephaint: Technically it isn't necessary because we catch it earlier but mypy won't shutup if I don't add it here.....

hierarchicalforecast/methods.py
elephaint: Thanks!

.github/workflows/ci.yml
```diff
@@ -6,57 +6,29 @@ on:
...
+          python-version: ${{ matrix.python-version }}
 
       - name: Install pip requirements
-        run: pip install ".[dev]"
+        run: pip install uv && uv pip install --system ".[dev]"
 
-      - name: Run tests
-        run: nbdev_test
+      - name: Tests
+        run: nbdev_test --do_print --timing --n_workers 0
```
jmoralez, line 34: I think we can run the tests in parallel here (by not setting `n_workers`). I didn't want to do it in the nixtla repo because it calls an API, but here it just does everything locally.

hierarchicalforecast/utils.py
```diff
@@ -476,12 +475,12 @@ def quantiles_to_outputs(quantiles:Iterable[float]):
...
 # given input array of sample forecasts and inptut quantiles/levels, 
 # output a Pandas Dataframe with columns of quantile predictions
-def samples_to_quantiles_df(samples:np.ndarray, 
-                            unique_ids:Iterable[str], 
-                            dates:Iterable, 
-                            quantiles:Optional[Iterable[float]] = None,
-                            level:Optional[Iterable[int]] = None, 
-                            model_name:Optional[str] = "model"):
+def samples_to_quantiles_df(samples: np.ndarray, 
+                            unique_ids: List[str],
```
jmoralez, line 479: I think these are better as `collections.abc.Sequence`

jmoralez (dismissed):
Minor non-blocking comments

.github/workflows/ci.yml
elephaint: Ok - I figured because the CI in NeuralForecast also has `n_workers 0`, should we change that one too then? It would help a lot in speeding up the tests.

jmoralez: Yeah sure we should try it. I see I changed it to zero in #829 but I think that was because I was getting segfaults and couldn't see the errors, before that it was 1 but if it works with multiprocessing then let's go for it.

hierarchicalforecast/utils.py
elephaint: Thanks - I looked it up and it's because it's a slightly more generic container?

jmoralez (approved):

PR #303, opened 2024-10-25
[FIX] Restructuring

elephaint:
- [x] Put source files in src folder
- [x] Restructure methods adding Sparse methods to docs
- [x] Update CI to improve speed of tests
- [x] Update model performance tests to test more models
- [x] Docs only update on release or workflow dispatch, not on push to main
- [x] Move Numba functions to utils, perform eager compilation and micro optimization of lasso function.
- [x] Use `double` dtype everywhere; reconcilers are sensitive to low(er) precision, for now unify to `double` and maybe in the future turn everything to `single`

Lasso (ERM-reg / ERM-reg_bu) micro optimization before and after (the method is still slow as **** ):

Screenshot 2024-10-29 104009

Screenshot 2024-10-29 103932


.github/release-drafter.yml
```diff
@@ -3,16 +3,14 @@ tag-template: 'v$NEXT_PATCH_VERSION'
 categories:
   - title: 'New Features'
     label: 'feature'
-  - title: 'Breaking Change'
-    label: 'breaking change'
```
jmoralez, line 7: Why were these removed?

.github/workflows/python-publish.yml
```diff
@@ -28,8 +28,8 @@ jobs:
           python-version: '3.x'
       - name: Install dependencies
         run: |
-          python -m pip install --upgrade pip
-          pip install build
+          python -m pip install uv
+          uv pip install build
```
jmoralez, line 32: I'm not sure this will work due to the venv. Since it's just a single package it's ok to leave it with pip. You could also use setup-uv and `uvx build` instead

action_files/test_models/src/data.py
```diff
@@ -35,9 +35,9 @@ def get_data():
     spec = [
         ['Country'],
         ['Country', 'State'], 
-        ['Country', 'Purpose'], 
+        # ['Country', 'Purpose'],
```
jmoralez, line 38: Why were these removed?

action_files/test_models/src/evaluation.py
```diff
@@ -32,7 +31,7 @@ def evaluate():
 
     eval_tags = {}
     eval_tags['Total'] = tags['Country']
-    eval_tags['Purpose'] = tags['Country/Purpose']
+    # eval_tags['Purpose'] = tags['Country/Purpose']
```
jmoralez, line 34: Why was this removed?

action_files/test_models/src/models.py
```diff
@@ -1,13 +1,15 @@
 import os
-import time
+# import time
```
jmoralez, line 2: If these are not used it's better to remove them

hierarchicalforecast/core.py
```diff
@@ -244,17 +244,16 @@ def reconcile(self,
                 S_for_sparse = sparse.csr_matrix(S_df.sparse.to_coo())
             except AttributeError:
                 warnings.warn('Using dense S matrix for sparse reconciliation method.')
-                S_for_sparse = S_df.values.astype(np.float32)
+                S_for_sparse = S_df.values.astype(np.float64)
```
jmoralez, line 247: I'd add `copy=False` in these in case the array is already double

hierarchicalforecast/utils.py
```diff
@@ -557,3 +565,202 @@ def samples_to_quantiles_df(samples: np.ndarray,
...
+                W[i, j] = W[j, i] = shrinkage * W[i, j]
+            else:
+                W[i, j] = W[j, i] = max(W[i, j], mint_shr_ridge)
+
+    return W
+
+# %% ../nbs/src/utils.ipynb 47
+# Lasso cyclic coordinate descent
+@njit("Array(float64, 1, 'C')(Array(float64, 2, 'C'), Array(float64, 1, 'C'), float64, int64, float64)", nogil=NUMBA_NOGIL, cache=NUMBA_CACHE, fastmath=NUMBA_FASTMATH, error_model="numpy")
+def _lasso(X: np.ndarray, y: np.ndarray,
```
jmoralez, line 728: Is this function working as expected? I get either very big numbers of zeros.

.github/release-drafter.yml
elephaint: Never mind, added them back.

action_files/test_models/src/data.py
elephaint: Hierarchy needs to be strictly hierarchical

action_files/test_models/src/evaluation.py
elephaint: Hierarchy needs to be strictly hierarchical

hierarchicalforecast/core.py
elephaint: Good point

hierarchicalforecast/utils.py
elephaint: This is a very valuable comment, thanks - this led to a rabbit hole uncovering that ERM-reg basically doesn't work at all, and it's also never tested (and was never tested before) - our unit test only tests ERM-reg_bu which does work. I looked back into old PRs and I don't think ERM-reg has ever worked.

To answer your questions:
- Zeros can be expected due to nan values present in the input data that ERM can't handle (I can think of how to handle these, but since ERM is essentially imho a fairly useless method as it's so computationally expensive I'm ok with this behaviour).
- If you'd perform a test without nan values you should see a P that's not full of zeros. 

I'll add a test for this and see if I can fix ERM-reg

elephaint: Btw these things are the reason I added much more methods to the model_performance test, as our testing is very limited. But I think I'll add even more tests, seeing these kinds of bugs.

elephaint: ERM-reg is basically garbage, I think we should leave it like this for now (old & new give same results)

elephaint:
Test for lasso equivalence of methods (removed it from the code to keep it clean):

```
#| hide
# Lasso test old vs new
@njit
def _lasso_old(X: np.ndarray, y: np.ndarray, 
          lambda_reg: float, max_iters: int = 1_000,
          tol: float = 1e-4):
    # lasso cyclic coordinate descent
    n, feats = X.shape
    norms = (X ** 2).sum(axis=0)
    beta = np.zeros(feats, dtype=np.float32)
    beta_changes = np.zeros(feats, dtype=np.float32)
    residuals = y.copy()

    for it in range(max_iters):
        for i, betai in enumerate(beta):
            # is feature is close to zero, we 
            # continue to the next.
            # in this case is optimal betai= 0
            # print(beta)

            if abs(norms[i]) < 1e-8:
                continue
            xi = X[:, i]
            #we calculate the normalized derivative
            rho = betai + xi.flatten().dot(residuals) / norms[i] #(norms[i] + 1e-3)
            #soft threshold
            beta[i] = np.sign(rho) * max(np.abs(rho) - lambda_reg * n / norms[i], 0.)#(norms[i] + 1e-3), 0.)
            beta_changes[i] = np.abs(betai - beta[i])
            if beta[i] != betai:
                residuals += (betai - beta[i]) * xi
        if max(beta_changes) < tol:
            break
    #print(it)
    return beta


S = np.array([[1., 1., 1., 1.],
              [1., 1., 0., 0.],
              [0., 0., 1., 1.],
              [1., 0., 0., 0.],
              [0., 1., 0., 0.],
              [0., 0., 1., 0.],
              [0., 0., 0., 1.]])
h = 2
_y = np.array([10., 5., 4., 2., 1.])
y_bottom = np.vstack([i * _y for i in range(1, 5)])
y_hat_bottom_insample = np.roll(y_bottom, 1)
y_hat_bottom = np.vstack([i * np.ones(h) for i in range(1, 5)])
idx_bottom = [3, 4, 5, 6]

y_hat=S @ y_hat_bottom
y_insample=S @ y_bottom
idx_bottom=idx_bottom

n_hiers, n_bottom = S.shape

for method in ["reg", "reg_bu"]:
    for lambda_reg in [None, 0.1, 10, 1000]:
        for with_nans in [True, False]:
            y_hat_bottom_insample = np.roll(y_bottom, 1)
            if with_nans:
                y_hat_bottom_insample[:, 0] = np.nan
            y_hat_insample=S @ y_hat_bottom_insample
            y_insample=S @ y_bottom

            nan_idx = np.isnan(y_hat_insample).any(axis=0)
            y_insample = y_insample[:, ~nan_idx]
            y_hat_insample = y_hat_insample[:, ~nan_idx]
            h = min(y_hat.shape[1], y_hat_insample.shape[1])
            y_hat_insample = y_hat_insample[:, -h:] # shape (h, n_hiers)
            y_insample = y_insample[:, -h:]

            if method == 'reg':
                X = np.kron(S, y_hat_insample.T)
                z = y_hat_insample.reshape(-1)

                if lambda_reg is None:
                    lambda_reg = np.max(np.abs(X.T.dot(z)))
                else:
                    lambda_reg = lambda_reg            

                beta_old = _lasso_old(X, z, lambda_reg)
                beta_new = _lasso(X, z, lambda_reg, max_iters=1000, tol=1e-4) 
                np.testing.assert_allclose(beta_new, beta_old, atol=1e-5, rtol=1e-3)

            if method == 'reg_bu':
                X = np.kron(S, y_hat_insample.T)
                Pbu = np.zeros_like(S)
                Pbu[idx_bottom] = S[idx_bottom]
                z = y_hat_insample.reshape(-1) - X @ Pbu.reshape(-1)
                    
                if lambda_reg is None:
                    lambda_reg = np.max(np.abs(X.T.dot(z)))
                else:
                    lambda_reg = lambda_reg             

                beta_old = _lasso_old(X, z, lambda_reg)
                beta_new = _lasso(X, z, lambda_reg, max_iters=1000, tol=1e-4) 
                np.testing.assert_allclose(beta_new, beta_old, atol=1e-5, rtol=1e-3)
```

action_files/test_models/src/data.py
jmoralez: I suggest we remove them instead of commenting them out.

action_files/test_models/src/evaluation.py
jmoralez: I suggest we remove them instead of commenting them out.

action_files/test_models/src/models.py
```diff
@@ -17,8 +19,27 @@ def main():
...
                    MinTrace(method='mint_shrink'),
-                   MinTrace(method='ols')]
+                   MinTraceSparse(method='ols'),
+                   MinTraceSparse(method='wls_struct'),
+                   MinTraceSparse(method='wls_var'),
+                   OptimalCombination(method='ols'),
+                   OptimalCombination(method='wls_struct'),
+                   ERM(method='closed'),
+                #    ERM(method='reg'), # This is so insanely slow that we don't run it
+                #    ERM(method='reg_bu'), # This is so insanely slow that we don't run it
```
jmoralez, line 41: If these methods don't work we may as well remove them, people may be trying them out in a grid search or similar and just wasting time.

jmoralez (approved):
Non blocking comments.

action_files/test_models/src/evaluation.py
elephaint: Yeah, leaving these here as future-me wants to create additional tests when I'm introducing temporal reconciliation techniques, I probably want to try out strict- and non-strict hierarchical stuff then too.

For now my rationale was to get a reasonable test on every existing method

action_files/test_models/src/models.py
elephaint: Agree, I'll be adding a deprecation notice in a future PR (or e.g. an exception when people try to run it with more than 10 timeseries)

PR #305, opened 2024-10-31
[FEAT] Add polars support

elephaint:
Adds support for Polars via [Narwhals](https://github.com/narwhals-dev/narwhals). 

The extension to all frameworks supported under Narwhals should hereafter be (nearly) trivial.

## Breaking
PR removes support for `NIXTLA_ID_AS_COL`, i.e., hierarchicalforecast from now on will return everything as a flat table, without an index in case of a Pandas DataFrame.

## Open items
- [x] Convert core.ipynb
- [x] Convert evaluation.ipynb
- [x] Convert methods.ipynb
- [x] Convert probabilistic_methods.ipynb
- [x] Convert utils.ipynb
- [x] Add tutorial on using with Polars
- [x] Re-run existing examples in the docs to check correct workings & adapt for deprecation of unique_id as index

### Fixes / better solutions required
- [x] Fix performance regression in aggregate function
- [x] Fix .agg function not allowing multiple aggregations of the same column (needed for exogenous)
- [x] Fix how to deal with unique_id being index everywhere but not for Polars frames
- [x] Fix plotting in utils - it now just converts Polars to Pandas when plotting, that isn't nice
- [x] Fix evaluation not aggregating correctly

## Tests
- [x] Add equivalence tests to unit tests for Polars
- [x] Inspect performance regression from adding Polars support

## Performance evaluation

Reconcilers seem to run somewhat faster after this PR. Reconciler0 = MinTShrink, Reconciler1 = BottomUp

This PR:

Screenshot 2024-11-12 232320

Current main:

Screenshot 2024-11-12 232328

Test script for PR in `tests\test_benchmark.py`, run with `pytest tests\test_benchmark.py::test_reconciler -v -s --benchmark-min-rounds=20 --disable-warnings`


hierarchicalforecast/utils.py
```diff
@@ -171,49 +106,41 @@ def aggregate(
...
     bottom = spec[-1]
     aggs = []
     tags = {}
     # Prepare the aggregation dictionary
-    agg_dict = {
-        "y": ("y", "sum")
-    }
-
-
+    agg_dict = {"y": "sum"}
```
elephaint, line 140: This should be fixed; performing the aggregation with ufp doesn't work with a dictionary

elephaint: Fixed

MarcoGorelli:
Hey @elephaint , thanks for trying out Narwhals! We can prioritise pivot as that would be helpful to you

We'd previously run into issues with Polars pre1.0 inconsistencies, but from this PR it looks you only need the 1.0+ behaviour, right?

Is there anything else we could do in Narwhals to help you out here?

elephaint:
> Hey @elephaint , thanks for trying out Narwhals! We can prioritise pivot as that would be helpful to you
> 
> We'd previously run into issues with Polars pre1.0 inconsistencies, but from this PR it looks you only need the 1.0+ behaviour, right?
> 
> Is there anything else we could do in Narwhals to help you out here?

Thanks! For this PR / repo I think indeed at the moment the only thing I miss is pivot - which I kind of get as it's an expensive op (as most fast frameworks are column based). 

Another minor thing that would be great is ability to ascertain datatypes at the op level without casting afterwards. For example, doing `.to_dummies()` with Polars will result in an Uint8 but with Pandas in a Float64. I know it can be done by subsequently casting over all the generated columns, but not sure this is the most efficient way when 1000+ columns are generated at once through the `.to_dummies()` command. But then again, this is minor, in this PR I settled on not using dummies. 

A thing that I found a bit unintuitive is that Narwhals seems to keep track of the order of a Pandas Integer Index even after narwhallifying it - I didn't expect that. Again minor thing, because I just need to remember to do `.maybe_reset_index()` after every op that potentially changes the order. 

But awesome library, keep up the good work! For this PR I wanted to play with it for Polars support, maybe later I'll try other frameworks - should almost work out of the box now (save for the pivot op).

MarcoGorelli:
Thanks so much for your feedback! 🙏 

- `pivot`: we're revisiting [the PR](https://github.com/narwhals-dev/narwhals/pull/546) and will try to make a release shortly
- `to_dummies`: we've addressed that as part of [this PR](https://github.com/narwhals-dev/narwhals/pull/1040) so that the output dtype is consistently Int8, we'd overlooked the dtype difference the first time round. Unsigned integers can be a bit unexpected when working with pandas, so Int8 seemed like a good compromise. Would this work for you?
- fragmentation warning: technically the warning is coming from pandas, but  it's really useful that you've reported this to us - I'll take a look to see if there's anything we're doing that's causing it (and whether it's an issue)

elephaint:
> Thanks so much for your feedback! 🙏
> 
> * `pivot`: we're revisiting [the PR](https://github.com/narwhals-dev/narwhals/pull/546) and will try to make a release shortly

Wow, awesome. I'll have a go at it once released.

> * `to_dummies`: we've addressed that as part of [this PR](https://github.com/narwhals-dev/narwhals/pull/1040) so that the output dtype is consistently Int8, we'd overlooked the dtype difference the first time round. Unsigned integers can be a bit unexpected when working with pandas, so Int8 seemed like a good compromise. Would this work for you?

Yes, perfect! I think in general what I'd expect (in an ideal world 🙈 - I know this is insanely hard to get perfect) from Narwhals is output equivalence up to DataFrame type (i.e. everything is the same, except the 'container' of the data - e.g. dtypes, row order, column order, column names, precision, nulls, nans, infs, categories are all similar). 

> * fragmentation warning: technically the warning is coming from pandas, but  it's really useful that you've reported this to us - I'll take a look to see if there's anything we're doing that's causing it (and whether it's an issue)

👍

action_files/test_models/src/data.py
```diff
@@ -42,24 +42,16 @@ def get_data():
...
-    Y_test_df = Y_test_df.set_index('unique_id')
-    Y_train_df = Y_train_df.set_index('unique_id')
-
-    sf = StatsForecast(df=Y_train_df,
-                       models=[ETS(season_length=4, model='ZZA')],
+    sf = StatsForecast(models=[AutoETS(season_length=4, model='ZZA')],
                        freq='QS', n_jobs=-1)
-    Y_hat_df = sf.forecast(h=8, fitted=True)
-    Y_fitted_df = sf.forecast_fitted_values()
+    Y_hat_df = sf.forecast(df=Y_train_df, h=8, fitted=True).reset_index()
```
jmoralez, line 52: I recommend instead setting the `NIXTLA_ID_AS_COL` env variable. Can be either in the CI job or in this script. That way you won't have to reset the index anywhere.

hierarchicalforecast/core.py
```diff
@@ -5,18 +5,19 @@
...
 # %% ../nbs/src/core.ipynb 4
 import re
-import gc
 import time
 import copy
+
 from .methods import HReconciler
+from .utils import pivot
 from inspect import signature
+from narwhals.typing import FrameT
```
jmoralez, line 14: I don't think `FrameT` is the right type here, TypeVars are useful when you want to indicate that the output type will be the same as one if the inputs, e.g. `def f(df: FrameT) -> FrameT`, that will tell type checkers that if you passed a pandas dataframe as `df`, you will get a pandas dataframe as output.

I believe what should be used here is either `Frame` or `IntoDataFrame`, depending on the case ([reference](https://narwhals-dev.github.io/narwhals/api-reference/typing/))

hierarchicalforecast/core.py
```diff
@@ -49,27 +55,27 @@ def _reverse_engineer_sigmah(Y_hat_df, y_hat, model_name):
...
-    drop_cols = ['ds']
-    if 'y' in Y_hat_df.columns:
-        drop_cols.append('y')
+    drop_cols = [time_col]
+    if target_col in Y_hat_df.columns:
+        drop_cols.append(target_col)
     if model_name+'-median' in Y_hat_df.columns:
         drop_cols.append(model_name+'-median')
-    model_names = Y_hat_df.drop(columns=drop_cols, axis=1).columns.to_list()
+    model_names = Y_hat_df.drop(drop_cols).columns
```
jmoralez, line 63: This will be very expensive for pandas dataframes, because a copy of the dataframe with the columns removed is made, only to extract the column names.
```suggestion
    model_names = [c for c in Y_hat_df.columns if c not in drop_cols]
```

hierarchicalforecast/core.py
```diff
@@ -49,27 +55,27 @@ def _reverse_engineer_sigmah(Y_hat_df, y_hat, model_name):
...
     pi_model_names = [name for name in model_names if ('-lo' in name or '-hi' in name)]
     pi_model_name = [pi_name for pi_name in pi_model_names if model_name in pi_name]
     pi = len(pi_model_name) > 0
 
-    n_series = len(Y_hat_df.index.unique())
+    n_series = len(Y_hat_df[id_col].unique())
```
jmoralez, line 68:
```suggestion
    n_series = Y_hat_df[id_col].n_unique()
```

hierarchicalforecast/core.py
```diff
@@ -97,99 +103,142 @@ def __init__(self,
...
-
-            if Y_df is not None:
-                Y_df = Y_df.reset_index()
-                Y_df.unique_id = Y_df.unique_id.astype('category')
-                Y_df.unique_id = Y_df.unique_id.cat.set_categories(S_df.index)
-                Y_df = Y_df.sort_values(by=['unique_id', 'ds'])
-                Y_df = Y_df.set_index('unique_id')
-
-            S_df.index = pd.CategoricalIndex(S_df.index, categories=S_df.index)
+        # TODO: This is now a bit slow as we always sort.
```
jmoralez, line 126: We should deprecate the `sort_df` argument then

hierarchicalforecast/core.py
```diff
@@ -97,99 +103,142 @@ def __init__(self,
...
+        S_df_nw = S_df_nw.with_columns(nw.from_dict({f"{id_col}_id": np.arange(len(S_df_nw))}, 
+                                                native_namespace=self.native_namespace)[f"{id_col}_id"])
+        Y_hat_df_nw = Y_hat_df_nw.join(S_df_nw[[id_col, f"{id_col}_id"]], on=id_col, how='left')
+        Y_hat_df_nw = Y_hat_df_nw.sort(by=[f"{id_col}_id", time_col])
+        Y_hat_df_nw = Y_hat_df_nw.drop(f"{id_col}_id")
+
+        if Y_df is not None:
+            Y_df_nw = nw.from_native(Y_df)
```
jmoralez, line 134: It's clearer to define this here
```suggestion
        if Y_df is None:
            Y_df_nw = None
        else:
            Y_df_nw = nw.from_native(Y_df)
```

hierarchicalforecast/core.py
```diff
@@ -97,99 +103,142 @@ def __init__(self,
...
-        if not len(Y_hat_df[model_names].select_dtypes(include='number').columns) == len(Y_hat_df[model_names].columns):
-            raise Exception('`Y_hat_df`s columns contain non numeric types')
-            
-        #Ensure no null values
-        if Y_hat_df[model_names].isnull().values.any():
-            raise Exception('`Y_hat_df` contains null values')
-        
-        pi_model_names = [name for name in model_names if ('-lo' in name or '-hi' in name or '-median' in name)]
-        model_names = [name for name in model_names if name not in pi_model_names]
+        assert Y_hat_df_nw[model_names].columns == Y_hat_df_nw[model_names].select(nw.selectors.numeric()).columns, "All models in `Y_hat_df` must contain numeric values."
```
jmoralez, line 161: This also looks like an expensive operation in pandas just to get the dtypes. Maybe can be done with `Y_hat_df_nw.schema` instead.
Also please use exceptions instead of assertions.

hierarchicalforecast/core.py
```diff
@@ -97,99 +103,142 @@ def __init__(self,
...
+
+        # Ensure no null values
+        assert not Y_hat_df_nw[model_names].select(nw.all().is_null().any()).to_numpy().any(), "`Y_hat_df` contains null values. Make sure no column in `Y_hat_df` contains null values."
```
jmoralez, line 164: This could potentially use a lot of memory, might be slower but it will use a lot less memory to do column wise or

hierarchicalforecast/core.py
```diff
@@ -97,99 +103,142 @@ def __init__(self,
...
         # Check Y_hat_df\S_df series difference
-        S_diff  = len(S_df.index.difference(uids))
-        Y_hat_diff = len(Y_hat_df.index.difference(S_df.index.unique()))
-        if S_diff > 0 or Y_hat_diff > 0:
-            raise Exception(f'Check `S_df`, `Y_hat_df` series difference, S\Y_hat={S_diff}, Y_hat\S={Y_hat_diff}')
+        # TODO: this logic should be method specific
+        S_diff = set(S_df_nw[id_col]) - set(Y_hat_df_nw[id_col])
+        Y_hat_diff = set(Y_hat_df_nw[id_col]) - set(S_df_nw[id_col])
+        if S_diff:
+            raise Exception(f'There are unique_ids in S_df that are not in Y_hat_df: {S_diff}')
```
jmoralez, line 182: These should be ValueErrors

hierarchicalforecast/core.py
```diff
@@ -97,99 +103,142 @@ def __init__(self,
...
+
+        # Same Y_hat_df/S_df/Y_df's unique_ids. Order is guaranteed by the sort_df flag.
+        # TODO: this logic should be method specific
+        unique_ids = Y_hat_df_nw[id_col].unique().to_numpy()
+        S_df_nw = S_df_nw.filter(nw.col(id_col).is_in(unique_ids))
+
+        return Y_hat_df_nw, S_df_nw, Y_df_nw, model_names
+
+    def _prepare_Y(self, 
+                          Y_df: FrameT,
```
jmoralez, line 202: I recommend setting `black_formatting = True` in the `settings.ini` to have consistent formatting in the source files.

elephaint:
> thanks! [it's released](https://github.com/narwhals-dev/narwhals/releases/tag/v1.13.5), and the fragmentation warning is hopefully addressed too

Awesome, thanks, I'll have a look at it

hierarchicalforecast/core.py
elephaint: Ah, very good point, fixed.

action_files/test_models/src/data.py
elephaint: Good suggestion

hierarchicalforecast/core.py
elephaint: THanks, forgot that drop copies, good point

hierarchicalforecast/core.py
elephaint: Agree

hierarchicalforecast/core.py
elephaint: Agreed, changed

hierarchicalforecast/core.py
elephaint: I love these detailed comments, thanks 👍

elephaint:
> thanks! [it's released](https://github.com/narwhals-dev/narwhals/releases/tag/v1.13.5), and the fragmentation warning is hopefully addressed too

Great work, I don't see the fragmentation messages anymore, and pivot works as expected. Thanks! 👍

hierarchicalforecast/core.py
```diff
@@ -90,170 +102,314 @@ class HierarchicalReconciliation:
...
+                nw.Int16,
+                nw.Int32,
+                nw.Int64,
+                nw.UInt8,
+                nw.UInt16,
+                nw.UInt32,
+                nw.UInt64,
+                nw.Float32,
+                nw.Float64,
+            ]:
```
jmoralez, line 198:
```suggestion
            if not Y_hat_df_nw.schema[model].is_numeric():
```

hierarchicalforecast/core.py
```diff
@@ -90,170 +102,314 @@ class HierarchicalReconciliation:
...
-        uids = Y_hat_df.index.unique()
+        # TODO: Complete y_hat_insample protection
+        model_names = [
+            name
+            for name in model_names
+            if not ("-lo" in name or "-hi" in name or "-median" in name)
+        ]
+        if intervals_method in ["bootstrap", "permbu"] and Y_df_nw is not None:
+            if not (set(model_names) <= set(Y_df_nw.columns)):
+                raise Exception(
```
jmoralez, line 215:
```suggestion
                raise ValueError(
```

hierarchicalforecast/core.py
```diff
@@ -90,170 +102,314 @@ class HierarchicalReconciliation:
...
 
-        uids = Y_hat_df.index.unique()
+        # TODO: Complete y_hat_insample protection
+        model_names = [
+            name
+            for name in model_names
+            if not ("-lo" in name or "-hi" in name or "-median" in name)
+        ]
+        if intervals_method in ["bootstrap", "permbu"] and Y_df_nw is not None:
+            if not (set(model_names) <= set(Y_df_nw.columns)):
```
jmoralez, line 214: Would be useful to save this in order to show the missing columns in the error message.

hierarchicalforecast/core.py
```diff
@@ -90,170 +102,314 @@ class HierarchicalReconciliation:
...
+                raise Exception(
+                    f"Check `Y_df` columns, {model_names} must be in `Y_df` columns."
+                )
+
+        # Assert S is an identity matrix at the bottom
+        S_df_nw_cols.remove(id_col)
+        if not np.allclose(
+            S_df_nw[S_df_nw_cols][-S_df_nw[S_df_nw_cols].shape[1] :],
+            np.eye(S_df_nw[S_df_nw_cols].shape[1]),
```
jmoralez, line 223:
```suggestion
            S_df_nw[S_df_nw_cols][-len(S_df_nw_cols) :],
            np.eye(len(S_df_nw_cols),
```

hierarchicalforecast/core.py
```diff
@@ -90,170 +102,314 @@ class HierarchicalReconciliation:
...
-                  seed: int = 0,
-                  sort_df: bool = True,
-                  is_balanced: bool = False,
-        ):
+        # TODO: this logic should be method specific
+        S_diff = set(S_df_nw[id_col]) - set(Y_hat_df_nw[id_col])
+        Y_hat_diff = set(Y_hat_df_nw[id_col]) - set(S_df_nw[id_col])
+        if S_diff:
+            raise ValueError(
+                f"There are unique_ids in S_df that are not in Y_hat_df: {S_diff}"
```
jmoralez, line 235: These could be a lot, might be worth using reprlib, e.g.
```suggestion
                f"There are unique_ids in S_df that are not in Y_hat_df: {reprlib.repr(S_diff)}"
```
same for the ones below.

hierarchicalforecast/core.py
```diff
@@ -90,170 +102,314 @@ class HierarchicalReconciliation:
...
+            },
         )
 
         any_sparse = any([method.is_sparse_method for method in self.reconcilers])
+        S_df_nw_cols_ex_id_col = S_df_nw.columns
+        S_df_nw_cols_ex_id_col.remove(id_col)
         if any_sparse:
+            if not is_pandas_Y_hat_df or not is_pandas_S_df:
+                raise ValueError(
+                    "You have one or more sparse reconciliation methods. Please convert `S_df` and `Y_hat_df` to a pandas DataFrame."
```
jmoralez, line 382: Does S have to be a dataframe? Can't we store it as a sparse matrix and just store the ids somewhere else? In my opinion we shouldn't even have dense methods, so not supporting sparse methods for polars seems inconvenient.

hierarchicalforecast/core.py
```diff
@@ -263,139 +419,203 @@ def reconcile(self,
...
+                Y_tilde_df = nw.concat(
+                    [
+                        Y_tilde_df,
+                        nw.from_dict(
+                            {recmodel_name: fcsts_model["mean"].flatten()},
+                            native_namespace=self.native_namespace,
+                        ),
+                    ],
+                    how="horizontal",
+                )
```
jmoralez, line 505: Does this not work?
```suggestion
                Y_tilde_df = Y_tilde_df.with_columns(**{recmodel_name: fcsts_model["mean"].flatten()})
```

jmoralez (changes requested):

setup.py
```diff
@@ -29,6 +29,11 @@
 min_python = cfg['min_python']
 lic = licenses.get(cfg['license'].lower(), (cfg['license'], None))
 dev_requirements = (cfg.get('dev_requirements') or '').split()
+polars_requirements = (cfg.get('polars_requirements') or '').split()
+all_requirements = [
```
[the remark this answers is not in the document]
elephaint: Yes, similar as we do in other libs

hierarchicalforecast/core.py
elephaint: At the moment, yes. The dense/sparse debate isn't something to solve in this PR. 

Polars doesn't support sparse.

setup.py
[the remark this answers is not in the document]
jmoralez: usually "all" means distributed as well.I don't think it makes sense having two extras that are the same

hierarchicalforecast/core.py
elephaint: Jup, better

setup.py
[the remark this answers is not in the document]
elephaint: Fine, whatever

hierarchicalforecast/core.py
elephaint: I don't understand what you mean?

hierarchicalforecast/core.py
```diff
@@ -90,170 +113,289 @@ class HierarchicalReconciliation:
...
+
         # Check input's validity
-        if intervals_method not in ['normality', 'bootstrap', 'permbu']:
-            raise ValueError(f'Unkwon interval method: {intervals_method}')
+        if intervals_method not in ["normality", "bootstrap", "permbu"]:
+            raise ValueError(f"Unknown interval method: {intervals_method}")
+
+        if self.insample or (intervals_method in ["bootstrap", "permbu"]):
+            if Y_nw is None:
+                raise Exception("You need to provide `Y_df`.")
```
jmoralez, line 164:
```suggestion
        if (self.insample or intervals_method in ["bootstrap", "permbu"]) and Y_df is None:
            raise ValueError("You need to provide `Y_df`.")
```

hierarchicalforecast/core.py
```diff
@@ -90,170 +113,289 @@ class HierarchicalReconciliation:
...
+        `id_col` : str='unique_id', column that identifies each serie.<br>
+        `time_col` : str='ds', column that identifies each timestep, its values can be timestamps or integers.<br>
+        `target_col` : str='y', column that contains the target.<br>
 
         **Returns:**<br>
-        `Y_tilde_df`: pd.DataFrame, with reconciled predictions.
+        `Y_tilde_df`: DataFrame, with reconciled predictions.
         """
+        # Raise deprecation warning for sort_df
+        _maybe_warn_sort_df(sort_df)
```
jmoralez, line 326: We could also remove this as part of the breaking changes

hierarchicalforecast/utils.py
```diff
@@ -7,42 +7,47 @@
...
-from numba import njit, prange
 import pandas as pd
+
+from narwhals.typing import Frame
+from numba import njit, prange
 from sklearn.preprocessing import OneHotEncoder
+from typing import Dict, List, Optional, Iterable, Union, Sequence
 
-plt.rcParams['font.family'] = 'serif'
+plt.rcParams["font.family"] = "serif"
```
jmoralez, line 21: This shouldn't be here
```suggestion
```

hierarchicalforecast/utils.py
```diff
@@ -77,384 +80,459 @@ def cov2corr(cov, return_std=False):
...
+        bottom_parts = bottom_value.split("/")
+        return "/".join(bottom_parts[i] for i in upper_idxs)
 
     return [join_upper(val) for val in bottom_values]
 
-# %% ../nbs/src/utils.ipynb 13
+# %% ../nbs/src/utils.ipynb 14
 def aggregate(
-    df: pd.DataFrame,
+    df: Frame,
```
jmoralez, line 96: you could use `FrameT` here to annotate the return as `tuple[FrameT, FrameT, dict]`

settings.ini
```diff
@@ -15,8 +15,9 @@ language = English
 custom_sidebar = True
 license = apache2
 status = 2
-requirements = numpy, numba, pandas, scikit-learn, quadprog, matplotlib, tqdm, intel-cmplr-lib-rt ; platform_system!="Darwin" and platform_machine=="x86_64"
-dev_requirements = datasetsforecast nbdev statsforecast>=1.0.0 requests scipy pre-commit ruff black pytest pytest-benchmark
+requirements = numpy, numba, pandas, scikit-learn>=1.2, quadprog, matplotlib, narwhals>=1.13.5, intel-cmplr-lib-rt ; platform_system!="Darwin" and platform_machine=="x86_64"
+dev_requirements = datasetsforecast nbdev statsforecast>=1.0.0 requests scipy pre-commit ruff black pytest pytest-benchmark polars
+polars_requirements = polars
```
jmoralez, line 20: to ensure numpy compatibility
```suggestion
polars_requirements = polars[numpy]
```

settings.ini
```diff
@@ -15,8 +15,9 @@ language = English
 custom_sidebar = True
 license = apache2
 status = 2
-requirements = numpy, numba, pandas, scikit-learn, quadprog, matplotlib, tqdm, intel-cmplr-lib-rt ; platform_system!="Darwin" and platform_machine=="x86_64"
-dev_requirements = datasetsforecast nbdev statsforecast>=1.0.0 requests scipy pre-commit ruff black pytest pytest-benchmark
+requirements = numpy, numba, pandas, scikit-learn>=1.2, quadprog, matplotlib, narwhals>=1.13.5, intel-cmplr-lib-rt ; platform_system!="Darwin" and platform_machine=="x86_64"
+dev_requirements = datasetsforecast nbdev statsforecast>=1.0.0 requests scipy pre-commit ruff black pytest pytest-benchmark polars
```
jmoralez, line 19:
```suggestion
dev_requirements = datasetsforecast nbdev statsforecast>=1.0.0 requests scipy pre-commit ruff black pytest pytest-benchmark
```

setup.py
```diff
@@ -29,6 +29,7 @@
 min_python = cfg['min_python']
 lic = licenses.get(cfg['license'].lower(), (cfg['license'], None))
 dev_requirements = (cfg.get('dev_requirements') or '').split()
+polars_requirements = (cfg.get('polars_requirements') or '').split()
```
jmoralez, line 32:
```suggestion
polars_requirements = (cfg.get('polars_requirements') or '').split()
dev_requirements.extend(polars_requirements)
```

jmoralez (dismissed):
non blocking comments

hierarchicalforecast/core.py
elephaint: Yeah, maybe better indeed to deprecate all these things in one go

hierarchicalforecast/core.py
elephaint: NB: `self.insample` is whether any method is insample, but whether Y_df is _not_ None varies across all methods being reconciled, so making your suggested change will lead to failures. (e.g. in the case where you have MinT with insample and a BottomUp)

The behavior is confusing, for now I've added a TODO comment to make it method-specific rather than partially across all methods, partially for a single method.

jmoralez (approved):

PR #309, opened 2024-11-27
[FEAT] Cross-temporal forecasting

elephaint:
## Temporal reconciliation strategy
- Aggregate temporally using `aggregate_temporal`, which is a minimal wrapper around `aggregate`, but in the temporal dimension, using a provided temporal `spec`
- Reconcile using our current reconcilers (except for in_sample reconcilers)

See the three new examples that showcase temporal reconciliation.

## Limitations
Temporal reconcilation (obviously) doesn't support in_sample reconcilers (as there are no insample residuals for the temporal aggregation in the test set, so we'd need an estimate for that - which seems good for future work)

## Open issues

- [x] ~~First example notebook doesn't work yet; main issue to solve is how the temporal unique ids can be matched back to the forecast df~~ Example works!
- [x] ~~Include an easier evaluate function for cross-temporal forecasts, it now requires to run a somewhat complex itertools double for-loop.~~ 
- [x] Unit tests for `aggregate_temporal`
- [x] Unit tests for `make_future_dataframe`
- [x] Add temporal-only example
- [x] Add temporal-probabilistic example
- [x] ~~ufe.time_features cannot handle non-timestamp ds columns~~ Input should be timestamp or integer
- [x] Add cross-temporal tags creation unit test
- [x] Fix failures on date features that don't naturally convert to dates
- [x] Add more unit tests on weird aggregation combinations


elephaint:
Seems like date features that are not year or days and don't generally easily convert to datetimes are still an issue to be solved 🤔

hierarchicalforecast/utils.py
```diff
@@ -238,7 +253,214 @@ def aggregate(
...
+        sparse_s=sparse_s,
+        id_col=id_time_col,
+        target_cols=target_cols,
+        time_col=id_col,
+    )
+    Y_nw = nw.from_native(Y_df)
+
+    # Add id_time_col column to Y_df
+    # This is a bit dirty but Narwhals does not support string splitting
+    timestamps = pd.to_datetime(
```
elephaint, line 351: This is problematic, as most date features don't convert nicely.

elephaint: Addressed, but open todo is to write more tests for this.

elephaint:
> ~Seems like date features that are not year or days and don't generally easily convert to datetimes are still an issue to be solved🤔~
> 
> ~This seems solved.~
> 
> Not quite yet. The aggregated ids are sorted lexicographically, which messes up the assignment of the dates

Fixed by maintaining order throughout the aggregation. Issues remaining:

- [ ] Assignment of bottom dates to aggregated levels is still iffy. Maybe a better direction is to exhaustively get the frequencies from the (limited set of) attributes, and then construct a daterange from a starting point and a number of periods, which are both known at each level.
- [ ] FIx handling of series of different lengths. Currently the aggregation can handle it but the timestamp assignment fails, it assigns `NaNs`

elephaint:
> > ~Seems like date features that are not year or days and don't generally easily convert to datetimes are still an issue to be solved🤔~
> > ~This seems solved.~
> > Not quite yet. The aggregated ids are sorted lexicographically, which messes up the assignment of the dates
> 
> Fixed by maintaining order throughout the aggregation. Issues remaining:
> 
> * [x]  Assignment of bottom dates to aggregated levels is still iffy. Maybe a better direction is to exhaustively get the frequencies from the (limited set of) attributes, and then construct a daterange from a starting point and a number of periods, which are both known at each level.
> * [x]  FIx handling of series of different lengths. Currently the aggregation can handle it but the timestamp assignment fails, it assigns `NaNs`
> * [x]  Add unit test on series of different lengths.

Assignment of bottom dates to aggregated levels by assigning the observed unique timestamps in a linspace where the step is determined by the number of steps in the temporal aggregation. Also not ideal, but best for the moment.

hierarchicalforecast/core.py
```diff
@@ -119,13 +122,55 @@ def _prepare_fit(
...
+            # Hence we also don't support bootstrap or permbu (rely on insample values)
+            if intervals_method in ["bootstrap", "permbu"]:
+                raise NotImplementedError(
+                    f"Temporal reconciliation is not supported for intervals_method=`{intervals_method}`."
+                )
+
+            missing_cols_temporal = set(
+                [id_col, time_col, target_col, id_time_col]
+            ) - set(Y_hat_nw_cols)
+            if len(missing_cols_temporal) > 0:
```
jmoralez, line 157: the pythonic way of checking if a collection is not empty is checking is truthiness, so this should be `if missing_cols_temporal:`

hierarchicalforecast/utils.py
```diff
@@ -82,6 +86,7 @@ def aggregate(
     id_col: str = "unique_id",
     time_col: str = "ds",
     target_cols: list[str] = ["y"],
+    maintain_order: bool = False,
```
jmoralez, line 89: This seems to complicate the implementation a bit, is there a use case that requires keeping the order?

hierarchicalforecast/utils.py
```diff
@@ -238,7 +287,245 @@ def aggregate(
...
+def aggregate_temporal(
+    df: Frame,
+    spec: list[list[str]],
+    freq: Union[str, int],
+    exog_vars: Optional[dict[str, Union[str, list[str]]]] = None,
+    sparse_s: bool = False,
+    id_col: str = "unique_id",
+    time_col: str = "ds",
+    id_time_col: str = "temporal_id",
+    target_cols: list[str] = ["y"],
```
jmoralez, line 300: I think we should make this immutable, e.g.

```python
from collections.abc import Sequence

...

target_cols: Sequence[str] = ("y",),  # one-tuple
```

hierarchicalforecast/utils.py
```diff
@@ -238,7 +287,245 @@ def aggregate(
...
+
+    # Check if ds column is a timestamp or integer, if not raise an error
+    df_nw = nw.from_native(df)
+    ds_pd = df_nw[time_col].to_pandas()
+    if not pd.api.types.is_datetime64_any_dtype(
+        ds_pd
+    ) and not pd.api.types.is_integer_dtype(ds_pd):
+        raise ValueError(
+            f"Check the dtype of '{time_col}', it must be a timestamp or integer"
+        )
```
jmoralez, line 354: We have a function in utils for this https://github.com/Nixtla/utilsforecast/blob/4823ba52aa34074dfbaf7ee83d0affcd10c98c0e/utilsforecast/validation.py#L47
so you could instead do:

```python
import utilsforecast.validation as ufv

...

df = ufv.ensure_time_dtype(df, time_col)
```

hierarchicalforecast/utils.py
elephaint: Yes, it's specifically required for the temporal reconciliation. Example. 

Suppose you add `month` as features. This will create a column with `1`, `2`, ...., `12` corresponding to the month of the year. If we group by those features to create a temporally aggregated timeseries on the `month` feature, and sort, we get a sorting like this `1`, `10`, `11`, ..., `9`. Thus, the aggregated timeseries will no longer have the correct temporal order (converting first to integer to overcome this doesn't cover all the cases that we face here btw). The issue is that we have no way of determining the correct temporal order after the aggregation, as there's no `ds` column anymore (and we need to add the `ds` column back afterwards in fact, which is the trickiest part of the code). The only way I could think of to overcome this issue is by maintaining the correct temporal order throughout the creation of all the aggregated timeseries.

hierarchicalforecast/utils.py
elephaint: Does this work when we have multiple targets? (which is the reason it's a list)

hierarchicalforecast/utils.py
elephaint: Awesome, great!

marcopeix:
I tried the functionality, and it is sound and easy to use. I think the workflow makes sense!

hierarchicalforecast/utils.py
jmoralez: yeah the comment about the one tuple was to see that there's an extra comma in there, but by annotating it with a sequence it can be ("y", "z"), ["y", "z"], etc

elephaint: Thanks - what's the benefit compared to having `list[str]`? I.e. what inputs do we prevent?

jmoralez: its not about the inputs but about having a mutable default argument, so for example if we modified target_cols inside the function with something like `target_cols.extend` or `target_cols.append`, the default value would be changed, so it's safer to have something immutable like a tuple.

elephaint: Sorry I still don't get it. If I do `list[str] = ["y"]` we can append or extend within the function without replacing the default value, i.e. `target_cols.append("u")` will yield `["y", "u"]`, which is fine. 

If I do `Sequence[str] = ["y",]` as default argument, I get the same results if I do `target_cols.append("u")`. With `Sequence[str]` I can still change/append/reassign `target_cols` inside the function. So what does immutable mean?

Do you have an example of where / how not having the Sequence gives an incorrect result?

jmoralez: The classic example is something like this:

```python
def f(arg_with_mutable_default=["y"]):
    arg_with_mutable_default.append("x")
    print(arg_with_mutable_default)

f()
# ['y', 'x']
f()
# ['y', 'x', 'x']
f()
# ['y', 'x', 'x', 'x']
```

so the safe approach is something like:

```python
def f(arg=("y",)):
    arg = list(arg)  # or assign some other name
    arg.append("x")
    print(arg)

f()
# ['y', 'x']
f()
# ['y', 'x']
```

elephaint: Whhaaaaatt that's crazy that Python works like that, why would you design it like that. Yeah it should definitely be immutable. How is muting default arguments like that even allowed..... 😮

elephaint: Thanks - fixed it.

jmoralez (approved):

Issue #329, opened 2025-02-16
`Y_df` parameter in `reconcile()` method requiring more columns than necessary when using TopDown or MiddleOut


macarw:
Hi! Thanks for this amazing library.

I was trying to reproduce the example notebook [**TourismSmall.ipynb**](https://github.com/Nixtla/hierarchicalforecast/blob/main/nbs/examples/TourismSmall.ipynb). In the example, you use BottomUp, TopDown with "forecast_proportions", and MiddleOut with "forecast_proportions", none of which require in-sample data.

However, when attempting to use the remaining TopDown and MiddleOut methods that are based on historical proportions ("average_proportions" and "proportion_averages"), and which do require in-sample data, I encountered the following errors:

`ColumnNotFoundError: The following columns were not found: [‘AutoARIMA’]`
or
`ColumnNotFoundError: The following columns were not found: [‘Naive']`

To clarify, the code is **identical** to the notebook, except that I added four additional reconcilers to the `reconcilers` list:
```
reconcilers = [
    # Used in the example notebook
    BottomUp(),
    TopDown(method="forecast_proportions"),
    MiddleOut(middle_level="Country/Purpose/State", top_down_method="forecast_proportions"),
    
    # Added for testing, raises the errors mentioned
    #TopDown(method=“average_proportions"),
    #TopDown(method=“proportion_averages"),
    #MiddleOut(middle_level=“Country/Purpose/State", top_down_method="average_proportions"),
    #MiddleOut(middle_level=“Country/Purpose/State", top_down_method="proportion_averages"),
]
```

In the notebook, the training dataset containing the columns `unique_id`, `ds` and `y` are passed as `Y_df`. This should be sufficient to compute historical proportions, yet these methods also requires `Y_df` to include the columns from `Y_hat_df` (e.g., `AutoARIMA` and `Naive`).

Why do we need to include **in-sample predictions** to use these methods, even though they should not rely on them?
I found a similar previous [issue](https://github.com/Nixtla/hierarchicalforecast/issues/253). 

I’d really appreciate any clarification.
Thanks in advance.

elephaint:
> Hi! Thanks for this amazing library.
> 
[...]

PR #330, opened 2025-02-17
[FIX] TopDown failing on combinations with other methods

elephaint:
Addresses both issues in #329:

- `TopDown` (also when used with `MiddelOut`) no longer requires insample predictions column to be present.
- `TopDown` failing when used in conjunction with other methods due to the insample attribute of reconciler being determined across all reconcilers rather than on a case-by-case basis.

Minor fixes:
- Docstring updates
- Protection added on some methods as a result of the new conditional check


marcopeix (approved):
Looks good to me!

PR #332, opened 2025-02-18
enh(methods): implement adjacency matrix

christophertitchen:
**Summary**

This PR implements functionality to construct the adjacency matrix for hierarchical or grouped time series that corresponds to the disaggregation process, ``utils._construct_adjacency_matrix``, and to check whether this adjacency matrix is strictly hierarchical, ``utils._is_strictly_hierarchcal``, for a sparse $S$.

We can use graph theory to massively improve the performance of the top-down disaggregation process for large structures, both for checking for a strictly hierarchical structure and calculating forecast proportions, the latter of which will be in a follow-up PR. **The new approach reduces the time taken to check the M5 data set from $`17.3 \, s`$ to $`19.9 \, ms`$.**

The (disaggregation) adjacency matrix, $A$, can be quickly constructed from the direct sum of the broadcasted Hadamard products of slices of a boolean $S$ belonging to successive levels, $\ell$, as shown below.

$`A = \begin{pmatrix} 0_{n_{a}} \quad\bigoplus_{i=1}^{K-1} S\left[\ell_{i} \right] \odot S\left[\ell_{i+1} \right]^{T} \\ 0_{n_{b}, n} \end{pmatrix}`$

where $K$ is the number of levels. This gives us the affinity, or simply the connectivity in this case as we only care about strictly hierarchical structures, between successive levels, which are diagonally stacked as a sparse block diagonal matrix and horizontally stacked with $0_{n_{a}}$ to give the truncated $`\left(n_{a} \times n\right)`$ $A$, which is all we need.

**Note:** $`S\left[\ell_{1} \right]`$ gives the root node, i.e., first row of the $S$, and $`S\left[\ell_{K} \right]`$ gives the $n_{b}$ leaf nodes, i.e., an identity matrix, and the other values of $i$ give the intermediate levels.

As the nodes in a strictly hierarchical disaggregation structure, except for the root node, should have exactly one incoming edge, we can simply check that the sum of each column in $A$ except for the first one is equal to $1$.

**Notes**

$A$ represents a disaggregation structure like the one below for the Wiki2 data set, which is a strictly hierarchical.

image

This can also be represented in a sparsity plot, like the one below.

image

However, we can clearly see that the Tourism data set below is not strictly hierarchical.

image


hierarchicalforecast/utils.py
```diff
@@ -44,6 +45,45 @@ def __exit__(self, exc_type, exc_value, traceback):
             )
 
 # %% ../nbs/src/utils.ipynb 7
+def _construct_adjacency_matrix(
+    S: sparse.csr_matrix, tags: dict[str, np.ndarray]
```
elephaint, line 49: Why only use this with TopDownSparse, not with MiddleOutSparse (I can't comment on the line in MiddleOutSparse)

hierarchicalforecast/utils.py
```diff
@@ -44,6 +45,45 @@ def __exit__(self, exc_type, exc_value, traceback):
...
+
+    """
+    # Get the nodes in each level.
+    l = list(tags.values())
+    # Get the number of aggregation nodes.
+    n_a = S.shape[0] - S.shape[1]
+    # Copy and cast the summing matrix to bool.
+    S = S.astype(bool)
+    # Precompute the transpose of the boolean summing matrix.
+    S_T = S.T
```
elephaint, line 73: Is this actually computed or is it just a view?

nbs/src/utils.ipynb
```diff
@@ -58,7 +59,16 @@
    "execution_count": null,
    "id": "3a501424-ebfd-403c-982e-280cb859bd0f",
    "metadata": {},
-   "outputs": [],
+   "outputs": [
```
elephaint, line 62: Can you clear the notebook?

nbs/src/utils.ipynb
```diff
@@ -130,6 +187,34 @@
...
+    "\n",
+    "    Returns\n",
+    "    -------\n",
+    "    bool\n",
+    "        `True` if strictly hierarchical, otherwise `False`.\n",
+    "\n",
+    "    \"\"\"\n",
+    "    return np.all(A.sum(axis=0).A1[1:] == 1)"
+   ]
+  },
```
elephaint, line 217: Can you include a unit test for this functionality? E.g. test equivalence to the dense method for a simple sparse structure.

elephaint (changes requested):
Great work! Few small comments 👍

hierarchicalforecast/utils.py
christophertitchen: Thanks for noticing this! I'll also change it to check for a strictly hierarchical structure just once when performing middle-out. 👍

hierarchicalforecast/utils.py
christophertitchen: Good question! It's not a view with modified strides in the usual sense.

The arrays (the indices, index pointer, and data) will share memory, but the transpose of a CSR matrix involves instantiating a CSC matrix, which is admittedly fast but orders of magnitude slower than transposing a numpy array for example. 😄

nbs/src/utils.ipynb
christophertitchen: Ah, I can see why this may be confusing. The tests for `is_strictly_hierarchical` is in `methods.ipynb`, so I chucked the tests `_is_strictly_hierarchical` in there too, along with tests for `_construct_adjacency_matrix`.

nbs/src/methods.ipynb
```diff
@@ -1344,30 +1395,21 @@
     "        level: Optional[list[int]] = None,\n",
     "        intervals_method: Optional[str] = None,\n",
     "    ) -> dict[str, np.ndarray]:\n",
-    "        \"\"\"Middle Out Sparse Reconciliation Method.\n",
```
elephaint, line 1347: Why delete this docstring?

elephaint (changes requested):

elephaint (changes requested):
Great! I think it's ready to go, one small comment re. docstring.

nbs/src/methods.ipynb
elephaint: You're right! Good point

elephaint (approved):
THanks!

PR #334, opened 2025-02-20
enh(methods): extremely fast forecast proportions

christophertitchen:
**Summary**

This PR uses the functionality introduced in #332 to construct a disaggregation adjacency matrix, $A$, to massively improve the performance of disaggregating with forecast proportions. **The new approach reduces the time taken to reconcile a strictly hierarchical subset of the M5 data set using forecast proportions from over $`1 \, hr`$ to around $`200 \, ms`$.**

We can use $A$ to find the siblings for each node by computing the Gram matrix, $`A^{T}A`$. $`\left( A^{T}A \right)\hat{y}_{h}`$ then gives us the $`\hat{y}_{h}`$ sums for each group of siblings, and hence $`\frac{\hat{y}_{h}}{\left( A^{T}A \right)\hat{y}_{h}}`$ gives us the _relative_ proportions for each node. 

Now that we have the _relative_ proportions for each node, we can isolate the _relative_ proportions for nodes along each leaf node's disaggregation pathway by computing the broadcasted Hadamard product, $`S^{T} \odot \frac{\hat{y}_{h}}{\left( A^{T}A \right)\hat{y}_{h}}`$, which gives us a sparse matrix, $P$.

$`P = S^{T} \odot \frac{\hat{y}_{h}}{\left( A^{T}A \right)\hat{y}_{h}}`$

The forecast proportions for each leaf node can then be obtained as the row-wise products of the strictly positive elements of $P$. The strictly positive elements can be efficiently obtained by converting $P$ to a LIL matrix and vertically stacking the lists.

$`p_{i} = \prod_{P_{ij}  \gt 0} P_{ij}`$

Finally, we can disaggregate the total forecast, $`\hat{y}_{0,h}`$, using $p$, and aggregate them with $S$ to generate the coherent reconciled forecasts.

$`\tilde{y}_{h} = S\left(p\hat{y}_{0,h} \right)`$


elephaint:
Great work again, I think if we merge #332 then this would only require a similar set of small changes (add test, make sure we add the functionality for all TopDownSparse and MiddleOutSparse)

hierarchicalforecast/methods.py
```diff
@@ -613,6 +613,79 @@ def _get_PW_matrices(
...
+                    "Top-down reconciliation requires strictly hierarchical structures."
+                )
+            A = A.astype(np.float64)
+            # As we may have zero sibling sums, replace any zeroes with eps.
+            y_hat[y_hat == 0.0] = np.finfo(np.float64).eps
+            # Calculate the relative proportions for each node.
+            with np.errstate(divide="ignore"):
+                P = y_hat / ((A.T @ A) @ y_hat)
+            # Set the relative proportion of the root node.
+            P[P == np.inf] = 1.0
```
elephaint, line 650: does `a / b` with `b = 0` always result in `np.inf`? E.g. also in the case of `0.0`, or `0.00000`. I guess so...

Otherwise we can also preinitialize P with ones (P = np.ones()) and fill where the denominator is not zero.

elephaint (changes requested):
Great work, small non-blocking comment and only a unit test is missing I guess.

hierarchicalforecast/methods.py
christophertitchen: I believe the only case where it would not be `np.inf` is if `a = 0` or `a = np.nan`, in which case it would be `np.nan`. 👍

christophertitchen:
@elephaint tests are failing due to a deprecation, but should be fixed by https://github.com/AnswerDotAI/execnb/pull/71.

christophertitchen:
> Great work, small non-blocking comment and only a unit test is missing I guess.

There are already unit tests to check the `forecast_proportions` output against both the expected output and the output of the non-sparse version. 😄

hierarchicalforecast/methods.py
elephaint: Thanks - how should we handle those cases then? `a = 0` + denominator also 0 seems a valid scenario. E.g. what happens in edge case all forecasts + all actuals are zeros?

elephaint (commented):
See comment on previous point, just making sure we capture the edge cases properly.

hierarchicalforecast/methods.py
elephaint: Thanks for the extremely good explanation!

elephaint (approved):

PR #335, opened 2025-02-23
Feature/date alignment error

janrth:
Because nan values can appear in average proportions and proportion averages without any warning, this PR creates a warning so the user is aware that the output consists of nan values only. @elephaint can you review please


elephaint:
Thanks for your PR!

I was wondering if this is the best / desired workflow. It raises a warning but the code still proceeds and `prop` may be NaN, leading to NaN results, no? 

I'd suggest one of these two alternatives:

1. Raise a ValueError instead of a warning: not allowing NaNs has the benefit that no unexpected behavior can occur.
2. Use `np.nanmean` to compute the proportions. In this case, we simply ignore the NaNs when computing the proportions. However again, unexpected behavior may occur, e.g. if there are a lot of NaNs. In this solution, I'd raise a warning.

What do you think? I think I prefer option 2 (`nanmean` for prop calculation + raising a warning)

janrth:
Makes a lot of sense! I will use nanmean calculation but also raise a warning if nan values are present!

janrth:
I hope the new changes are somewhat close to what you had in mind.

hierarchicalforecast/methods.py
```diff
@@ -405,16 +405,25 @@ def _get_PW_matrices(
...
-            prop = np.mean(y_btm, axis=1) / np.mean(y_top)
+            prop = np.nanmean(y_btm, axis=1) / np.nanmean(y_top)
         elif self.method == "forecast_proportions":
             raise NotImplementedError(
                 f"Fit method not implemented for {self.method} yet"
             )
         else:
             raise ValueError(f"Unknown method {self.method}")
 
+        if np.isnan(prop).any():
```
elephaint, line 418: There won't be any NaN now in prop, so the check should be on `y_btm` and `y_top`, i.e.

`if np.isnan(y_btm).any() or np.isnan(y_top).any()`

nbs/src/methods.ipynb
```diff
@@ -733,13 +733,22 @@
...
-    "            prop = np.mean(y_btm / y_top, axis=1)\n",
+    "            prop = np.nanmean(y_btm / y_top, axis=1)\n",
     "        elif self.method == \"proportion_averages\":\n",
-    "            prop = np.mean(y_btm, axis=1) / np.mean(y_top)\n",
+    "            prop = np.nanmean(y_btm, axis=1) / np.nanmean(y_top)\n",
     "        elif self.method == \"forecast_proportions\":\n",
     "            raise NotImplementedError(f\"Fit method not implemented for {self.method} yet\")\n",
     "        else:\n",
     "            raise ValueError(f\"Unknown method {self.method}\")\n",
+    "        \n",
```
elephaint, line 743: Can you include a unit test that tests the behavior in case there are NaNs present? You can look at the other tests in the `methods.ipynb`, it can be something relatively simple.

elephaint (changes requested):
Nice! Almost there 👍

hierarchicalforecast/methods.py
janrth: so true! I will correct it.

nbs/src/methods.ipynb
janrth: I first jumped directly on your other comment, but also added a unit test now. It looks slightly different than the other unit tests from this section, but I hope it is fine, but let me know if you want it different. I take the original values, add  a row of zeros and get the mean, which is my original value. Then I replace this zero row with nan values and because the row is zeros the results should be the same as my original values. I hope this makes sense.

janrth: I also saw that there is TopDownSparse class. Should we do the same also here if we have that PR open, or better a new one (if the change is desired here at all)?

nbs/src/methods.ipynb
```diff
@@ -1178,6 +1186,34 @@
...
    ]
   },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "#| hide\n",
+    "for method in [\"average_proportions\", \"proportion_averages\"]:\n",
```
elephaint, line 1196: Thanks for including the test! It fails, can you make sure it passes and produces the correct result?

You can try out locally by either simply running `methods.ipynb` or by doing the following in a terminal (from the base directory of the repo):

`nbdev_test --path nbs\src\methods.ipynb`

elephaint (changes requested):
Good work, only your included test fails, so if you can make that succeed I guess we're done....

nbs/src/methods.ipynb
janrth: ah, srry. i tested in a nb and then made copy paste error and used the sparse matrix in my test, which obviously didn't work. it should be fine now.

elephaint (approved):
Great, thanks! I'll merge once the tests pass (there's an issue with a dependency causing some failing tests - this is happening in all our PRs / repos - but we'll resolve in another PR).

elephaint (dismissed):

elephaint (approved):
Thanks again!

janrth:
lovely :)

PR #344, opened 2025-03-11
fix: ensure levels are sorted prior to adjacency matrix construction

### LINKED ISSUE #343: Strict hierarchy check fails if tags are not sorted
Hi,

My code started to break with the latest release 1.1.0 which changes the check for strict hierarchy. I started getting "Middle-out reconciliation requires strictly hierarchical structures." even though I had a strict hierarchy input.
I then realised that the issue was because my tags weren't sorted from top to bottom level, I don't think this requirement is mentioned explicitly and it can lead to unexpected behaviours.

I would suggest either:
1. adding an implicit sort everywhere tags are passed and the ordering is important, it should be possible to rely on the number of elements in each level to perform that

For example:
`tags = dict(sorted(tags.items(), key=lambda item: len(item[1])))`

2. adding an explicit requirement in the documentation and API reference and adding a check before relying on that assumption

Please let me know what you think, thanks!

christophertitchen:
This PR closes #343 by sorting the values in `tags` in `_construct_adjacency_matrix` prior to constructing the adjacency matrix.

### DIFF
--- a/hierarchicalforecast/utils.py
+++ b/hierarchicalforecast/utils.py
@@ -64,7 +64,7 @@ def _construct_adjacency_matrix(
 
     """
     # Get the nodes in each level.
-    l = list(tags.values())
+    l = sorted(tags.values(), key=lambda x: len(x))
     # Get the number of aggregation nodes.
     n_a = S.shape[0] - S.shape[1]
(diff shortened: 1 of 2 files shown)


elephaint (approved):
Good catch, LGTM

PR #362, opened 2025-04-10
fix: support disconnected graphs for temporal disaggregation

### LINKED ISSUE #360: TopDownSparse and MiddleOutSparse fail on temporal reconciliation
### What happened + What you expected to happen

The adjacency matrix calculation fails if the provided `S` does not contain a top level. Consequently, `TopDownSparse` and `MiddleOutSparse` fail on temporal reconciliation as these methods commonly do not have a top aggregation row.

The error can be reproduced by running the AustralianTourismTemporal example, using `TopDownSparse` and `MiddleOutSparse` as reconcilers.

### Versions / Dependencies

1.2.0

### Reproduction script

The error can be reproduced by running the AustralianTourismTemporal example, using `TopDownSparse` and `MiddleOutSparse` as reconcilers.

### Issue Severity

Medium: It is a significant difficulty but I can work around it.

christophertitchen:
This PR closes #360 by adding support for constructing the disaggregation adjacency matrix for disconnected graphs, which will commonly occur during temporal reconciliation due to the structure of the temporal $S$. As long as the temporal subgraphs are strictly hierarchical, the disaggregation will be correct.


elephaint:
Wow that's quick, thanks!

christophertitchen:
@elephaint I just realised that I changed everything apart from `forecast_proportions`, so I just made a change to support having multiple root nodes. Without the change, it would've disaggregated with the correct proportions, but always using the first element in $`\hat{y}_{h}`$. 

The $p$ vector of forecast proportions will sum to the number of root nodes, and these disaggregation proportions are now correctly distributed from each root node, such that the sum of each column in our truncated mapping matrix, $`S\left[\ell_{0} \right]^{T}  \odot \, p^{T}`$, is equal to one. 

Now that we have the truncated mapping matrix, it's straightforward to reconcile the forecasts. We can obtain the reconciled bottom-level forecasts, $`b_{h}`$, with $`\left(S\left[\ell_{0} \right]^{T}  \odot \, p^{T} \right) \hat{y}_{\ell_{0},h}`$, and finish the reconciliation with $`Sb_{h}`$.

elephaint:
> @elephaint I just realised that I changed everything apart from `forecast_proportions`, so I just made a change to support having multiple root nodes. Without the change, it would've disaggregated with the correct proportions, but always using the first element in y ^ h .
> 
> The p vector of forecast proportions will sum to the number of root nodes, and these disaggregation proportions are now correctly distributed from each root node, such that the sum of each column in our truncated mapping matrix, S [ ℓ 0 ] T ⊙ p T , is equal to one.
> 
> Now that we have the truncated mapping matrix, it's straightforward to reconcile the forecasts. We can obtain the reconciled bottom-level forecasts, b h , with ( S [ ℓ 0 ] T ⊙ p T ) y ^ ℓ 0 , h , and finish the reconciliation with S b h .

Nice! Great work. 

When I'm running MiddleOutSparse it still fails with the same error unfortunately, i.e.
image

`TopDown` and `TopDownSparse` are equal now 👍

elephaint:
Note that when I run [the example](https://nixtlaverse.nixtla.io/hierarchicalforecast/examples/australiandomestictourismtemporal.html) with the following settings:

image

MiddleOut and TopDown should produce the same results, which they nicely do:

image

But when I change to a proper middle level, e.g. 'semiannual':
image

MiddleOutSparse will error, but not the dense variant:

image

christophertitchen:
@elephaint I had a quick look and made a small change to setting the relative proportions of the root nodes—it should be good now.

elephaint:
> @elephaint I had a quick look and made a small change to setting the relative proportions of the root nodes—it should be good now.

Perfect, thanks! Works indeed, it now produces same results as the dense version:

image

PR #368, opened 2025-04-24
[FIX] TopDown with "forecast_proportions" should error when providing a non-strict hierarchy.

elephaint:
`TopDown(method="forecast_proportions")` should fail when provided a non-strict hierarchy. We capture this for `TopDownSparse`, but not for `TopDown`. This PR adds a strictness check to `TopDown` when used with `forecast_proportions`.

The error can be reproduced by running the Prison population example from the docs, which contains a non-strict hierarchy that **doesn't** fail when using `TopDown(method="forecast_proportions")` as reconciler, whereas it should fail. 

After this PR, a user is given the error message that `TopDown(method="forecast_proportions")` requires a strict hierarchy.


marcopeix (approved):
LGTM

PR #377, opened 2025-07-04
[FIX] Remove conda suggestions

elephaint:
Removes `conda` as package manager in favor of `uv`


elephaint:
> LGTM but strictly speaking we are not removing conda, as is still mentioned as an option in the README.

Yeah correct, it's indeed "preference". I'm also not taking it out of the publish scripts.

Issue #292, opened 2024-10-09
Better Documentation for combining different models (MLforecast + statsforecast for example)


jan-voids:
### Description

The given examples are working, but I find it hard to setup more advanced frameworks where I want to combine separate models, which I trained individually. Especially the usage of the correct indexing can be tricky.

elephaint:
I think you're right - what would such an example look like in your case?

Maybe we can start with an existing example and fit it to your needs? (i.e. using multiple models?)

jan-voids:
If we take the point reconciliation (tourism) example, it would be cool to show how a model from the mlforcast family on a lower level can be combined with statsforecast models from higher levels.

PR #378, opened 2025-07-06
[DOC] Adds example detailing the use of different forecasting methods when doing HF

nasaul:
Adds documentation suggested by #292


elephaint:
Thanks @nasaul. It's a good start. See my review comments; the example should be a new example, it should show the use of different types of forecasters and how one can use those.

nbs/examples/AustralianDomesticTourism.ipynb
```diff
@@ -699,6 +947,67 @@
...
+    "We'll be cleaning the columns names and the following function will help us do it in a more structured way:"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "def rename_evaluation_columns(\n",
```
elephaint, line 963: This is way too complicated, please remove this

nbs/examples/AustralianDomesticTourism.ipynb
```diff
@@ -852,7 +1309,7 @@
     "### MASE\n",
     "\n",
     "\n",
-    "The following table shows the performance measured using MASE across levels for each reconciliation method."
+    "The following table shows the performance measured using MASE across levels for each reconciliation method. Focusing only in AutoETS model"
```
elephaint, line 1312: Please include a qualitative evaluation across the example showcasing and evaluating the pros and cons of using different forecasting methods in the HF setup.

The issue is about using complex pipelines in HF, when one wants to combine multiple models _across different levels of the hierarchy_. None of the code demonstrates this. 

The idea is e.g. that one level is trained using AutoETS, one using LightGBM, one using MLP, etc. So not training the entire hierarchy using AutoETS or LightGBM. Thus, the code should include an explicit for loop over the hierarchy, where depending on the level of the hierarchy a different model is trained. The predictions should then be gathered across all the hierarchies.

nbs/examples/AustralianDomesticTourism-Multimodel.ipynb
```diff
@@ -0,0 +1,1732 @@
...
+    "The following cell computes the **base forecasts** for each time series in `Y_df` using the `ETS` model. Observe that `Y_hat_df` contains the forecasts but they are not coherent."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "Y_train_df = (\n",
```
elephaint, line 484: This helper function isn't needed, you can just iterate over the hierarchies by using the tags generated from the aggregate function.

nbs/examples/AustralianDomesticTourism-Multimodel.ipynb
```diff
@@ -0,0 +1,1732 @@
...
+    "from neuralforecast import NeuralForecast"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "def fit_predict_any_model(model: StatsForecast | MLForecast | NeuralForecast, df: pd.DataFrame, h:int):\n",
```
elephaint, line 531: This is a good helper function. It also shows how inconsistent our API is across libraries 😬 

Why is the `step_size` in `predict_insample` of NF `h`? (I think it should be the default value of `1`)

nbs/examples/AustralianDomesticTourism-Multimodel.ipynb
```diff
@@ -0,0 +1,1732 @@
...
+    "    return yhat, yfitted"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "def create_predictions(model, hierarchy: np.ndarray, df: pd.DataFrame, h: int=8) -> Tuple[pd.DataFrame, pd.DataFrame]:\n",
```
elephaint, line 552: Loop over the tags, and query the level using the tags.

nbs/examples/AustralianDomesticTourism-Multimodel.ipynb
```diff
@@ -0,0 +1,1732 @@
...
+     ]
+    }
+   ],
+   "source": [
+    "neural_model = NeuralForecast(\n",
+    "    models=[\n",
+    "        NBEATS(\n",
+    "            h=8,\n",
+    "            input_size=16,\n",
+    "            mlp_units=[[256, 256], [256, 256], [256, 256]],\n",
```
elephaint, line 663: I'd drop all non necessary arguments (ie basically everything except `h` and `input_size`). The example is about how to include different model types, so we don't need to unnecessarily clutter it with different hyperparameter choices.

nbs/examples/AustralianDomesticTourism-Multimodel.ipynb
```diff
@@ -0,0 +1,1732 @@
...
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "ml_model = MLForecast(\n",
+    "    models = {\n",
+    "        'gbm':HistGradientBoostingRegressor()\n",
+    "    }, \n",
+    "    freq='QS',\n",
+    "    target_transforms=[Differences([1, 4])],\n",
+    "    lags=[1, 2, 3, 4, 5, 6, 7, 8, 12],\n",
```
elephaint, line 615: Simplify, the example isn't about creating the best ML model + features, it's merely to showcase the use of different forecasters for each level in the hierarchy.

nbs/examples/AustralianDomesticTourism-Multimodel.ipynb
```diff
@@ -0,0 +1,1732 @@
...
+       "9  Australia/ACT 2016-04-01    568.944875"
+      ]
+     },
+     "execution_count": null,
+     "metadata": {},
+     "output_type": "execute_result"
+    }
+   ],
+   "source": [
+    "Y_hat_df_stats.head(10)"
```
elephaint, line 1066: I think the core logic is (still) incorrect as the dataframe shows forecasts for different levels with the same algorithm.

elephaint (changes requested):
Thanks Saul, made some good changes.

In addition to comments below, I'd suggest to read [the quickstart](https://nixtlaverse.nixtla.io/hierarchicalforecast/examples/tourismsmall.html) and [the introduction](https://nixtlaverse.nixtla.io/hierarchicalforecast/examples/introduction.html), so you get a better understanding of the different matrices `Y_df,` `S_df` and `tags` that are central to HF.

nbs/examples/AustralianDomesticTourism-Multimodel.ipynb
```diff
@@ -0,0 +1,630 @@
...
+    "\n",
+    "    elif isinstance(model, MLForecast):\n",
+    "        model.fit(df, fitted=True)\n",
+    "        yhat = model.predict(new_df=df, h=h)\n",
+    "        yfitted = model.forecast_fitted_values()\n",
+    "\n",
+    "    elif isinstance(model, NeuralForecast):\n",
+    "        model.fit(df=df, val_size=h)\n",
+    "        yhat = model.predict()\n",
+    "        yfitted = model.predict_insample(step_size=h).drop(columns=[\"cutoff\"])[[\"unique_id\", \"ds\", \"y\", \"NBEATS\"]]\n",
```
elephaint, line 307: Can you abstract the column selection from the model name? (so that there's no fixed `NBEATS` in here)

nbs/examples/AustralianDomesticTourism-Multimodel.ipynb
```diff
@@ -0,0 +1,630 @@
...
+    "This `fit_predict_any_model` function provides a unified interface for training and forecasting with models from `StatsForecast`, `MLForecast`, and `NeuralForecast`. By abstracting away the specific fit and predict methods of each library, it simplifies the process of applying diverse model types across different levels of a hierarchy. This utility streamlines the experimentation and implementation of multi-model hierarchical forecasting strategies, enhancing code readability and reusability."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "def fit_predict_any_model(model: StatsForecast | MLForecast | NeuralForecast, df: pd.DataFrame, h:int):\n",
```
elephaint, line 294: Technically StatsForecast | ML | NF can contain multiple models, and an individual model is part of the invoked class. So I'd use the plural `models` here, as SF | MF | NF commonly are containers for multiple models.

nbs/examples/AustralianDomesticTourism-Multimodel.ipynb
```diff
@@ -0,0 +1,630 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "%%capture\n",
+    "\n",
+    "stat_model = StatsForecast(models=[AutoETS(season_length=4, model='ZZA')], freq='QS', n_jobs=-1)\n",
```
elephaint, line 325: I'd use plural here too, `stats_models`, etc.

nbs/examples/AustralianDomesticTourism-Multimodel.ipynb
```diff
@@ -0,0 +1,630 @@
...
+    "The following cell makes the previous forecasts coherent using the `HierarchicalReconciliation` class. In this example we use `BottomUp` reconciler."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from hierarchicalforecast.methods import BottomUp\n",
```
elephaint, line 380: I'd use another reconciler too, because now we're throwing away the `stat_model` and `ml_model` in the reconciliation. E.g. also use `TopDown(forecast_proportions)` and `MinT(mint_shrink)`. Also, `Y_fitted_df` isn't necessary if we only stick to `BottomUp`, so then we don't even need to calculate it.

nbs/examples/AustralianDomesticTourism-Multimodel.ipynb
```diff
@@ -0,0 +1,630 @@
...
+       "      <td>mase</td>\n",
+       "      <td>1.42</td>\n",
+       "      <td>1.41</td>\n",
+       "    </tr>\n",
+       "  </tbody>\n",
+       "</table>\n",
+       "</div>"
+      ],
+      "text/plain": [
+       "     level metric  Base  BottomUp\n",
```
elephaint, line 599: What is the conclusion?

nbs/examples/AustralianDomesticTourism-Multimodel.ipynb
```diff
@@ -0,0 +1,630 @@
...
+       "3  Overall   mase  1.42      1.41"
+      ]
+     },
+     "execution_count": null,
+     "metadata": {},
+     "output_type": "execute_result"
+    }
+   ],
+   "source": [
+    "evaluation.query('metric == \"mase\"')"
```
elephaint, line 612: It's good to add a recap at the end briefly summarizing what we did in the example.

nbs/examples/AustralianDomesticTourism-Multimodel.ipynb
```diff
@@ -0,0 +1,630 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "In this section, we illustrate how to fit a different type of model for each level of the hierarchy. In particular, for each level, we will fit the following models:\n",
+    "\n",
+    "* **Country**: `AutoETS` model from `StatsForecast`.\n",
+    "* **State**: `HistGradientBoostingRegressor` model from `scikit-learn` through the `MLForecast` API.\n",
```
elephaint, line 254: **Country/State**

nbs/examples/AustralianDomesticTourism-Multimodel.ipynb
```diff
@@ -0,0 +1,630 @@
...
+    "* **Region**: `NBEATS` model from `NeuralForecast`."
```
elephaint, line 255: **Country/State/Region**

elephaint (changes requested):
Nice work @nasaul!

I think we're almost there 👍

nbs/examples/AustralianDomesticTourism-Multimodel.ipynb
nasaul: Here the problem is the order of the columns that differs from Stats or MLForecast. I think that it's better to remove this inside this function and when joining the data frames I can use a best_stats_model="AutoETS", etc and rename with that variables. Can be more useful when having lots of models. What do you think?

elephaint (approved):
Thanks @nasaul!

PR #379, opened 2025-07-07
[FIX] Deprecate S in favor of S_df

elephaint:
Deprecates `S` in `.reconcile` in favor of `S_df` to unify the API.


nbs/src/core.ipynb
```diff
@@ -377,8 +377,8 @@
     "\n",
```
marcopeix, line 377: <pre class="ql-syntax" spellcheck="false">Line #280.                    raise ValueError("You have one or more sparse reconciliation methods. Please convert <code>S</code> and <code>Y_hat_df</code> to a pandas DataFrame.")
</pre><p>I think the message should be "Please convert <code>S_df</code>..."</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Nixtla/hierarchicalforecast/pull/379/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='11'/>

PR #381, opened 2025-07-11
use `backend=` instead of deprecated `native_namespace=` in `nw.from_dict`

MarcoGorelli:
`native_namespace` in `nw.from_dict` is deprecated since Narwhals 1.27, which is already your minimum version

https://github.com/Nixtla/hierarchicalforecast/blob/434c411c475e8b5ce852943e018e77e16c198f05/settings.ini#L18

so is safe to do even for the oldest versions

For future reference, I'd also suggest using `narwhals.stable.v1` to guard against API changes https://narwhals-dev.github.io/narwhals/backcompat/

### DIFF
--- a/hierarchicalforecast/evaluation.py
+++ b/hierarchicalforecast/evaluation.py
@@ -486,7 +486,7 @@ class HierarchicalEvaluation:
                 },
                 **dict(zip(model_names, evaluation_np.T)),
             },
-            native_namespace=native_namespace,
+            backend=native_namespace,
         )
 
         evaluation = evaluation_nw.to_native()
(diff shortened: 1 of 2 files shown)


elephaint:
Thanks, makes sense!

Agree wrt stable API, but we need functions that are not in v1, so then it doesn't work, right?

elephaint (approved):

elephaint:
> thanks!
> 
> yup, true - do you remember which non-v1 functions you need?
> 
> (btw, we're close to having v2 🥳 )

I think it was the pivot function. Nice to see v2 coming! I want to start moving the rest of our libs to Narwhals too, but it'll be a big effort.

PR #385, opened 2025-07-23
[FIX] Add new doc to mint

elephaint:
Forgot to add the new doc from @nasaul to mint.json, causing it not to show up on the website.


deven367 (changes requested):
These changes would go in the `docs/mintlify/mint.json`, and the entry would be the name of the notebook (not ending in html)

nasaul:
I don't have privileges to do this in the same pull request but here's the solution. #392

PR #468, opened 2026-03-17
[FIX] Adds polars to extras

nasaul:
- [x]  Adds `polars` to optional-dependencies
- [x]  Automatically updates the version for mkdocstring-parser so we don't need to update the uv.lock with each correction of mkdocstring-parser


JQGoh (approved):
LGTM

PR #469, opened 2026-03-18
[FEAT] Support sparse S for any DataFrame type

elephaint:
Introduces `SMatrix`, a lightweight sparse wrapper for the hierarchical summing matrix S, replacing the previous approach of converting sparse matrices to dense DataFrames. When `aggregate(..., sparse_s=True)` is used, the S matrix now stays sparse throughout the pipeline — from construction through reconciliation, avoiding dense materialisation.

This also allows a full sparse pipeline when working with Polars.


tests/test_core.py
```diff
@@ -1008,3 +1008,183 @@ def test_non_negative_output(self):
...
+    sparse_data = sparse_grouped_data[lib]
+
+    assert isinstance(sparse_data["S_df"], SMatrix)
+
+    # Use both sparse and dense reconcilers
+    reconcilers = [
+        BottomUp(),
+        MinTrace(method='ols'),
+        MinTrace(method='mint_shrink'),
+    ]
```
marcopeix, line 1071: Aren't these all dense reconcilers? Maybe the comment on top is throwing me off?

marcopeix (commented):
Looks good, just see my small comment, I just wanna make sure it matches what you meant to do.

tests/test_core.py
elephaint: Yeah, you're right, forgot to add Sparse reconcilers here. It needs to work on both.

marcopeix (approved):
Very nice!

PR #492, opened 2026-07-25
Optimize bottom identity validation for summing matrices

janrth:
## Summary

Optimizes validation of the bottom identity block in summing matrices without changing the public reconciliation API or skipping validation.

- Materializes only the bottom `n_bottom × n_bottom` block for dense inputs.
- Uses a fast exact check for the common 0/1 identity case.
- Preserves the tolerant `np.allclose` behavior for user-built matrices with small floating-point differences.
- Handles pandas sparse columns whose fill value cannot be converted to COO.
- Caches successful identity checks for `SMatrix` instances.
- Marks sparse matrices produced by `aggregate` and `aggregate_temporal` as verified by construction.
- Disables identity-result caching after `to_sparse()` exposes the mutable backing matrix, ensuring subsequent mutations are detected.

No new argument is added to `HierarchicalReconciliation.reconcile`.

## Why

For large summing matrices, repeatedly constructing and comparing against a dense identity matrix adds noticeable runtime and memory overhead. The common case uses exact 0/1 matrices produced by the aggregation utilities, so it can
be validated more efficiently.

Repeated reconciliation with an unchanged `SMatrix` also no longer repeats the same sparse identity check.

## Safety

Bottom-identity validation remains enabled by default:

- Malformed dense and sparse summing matrices are rejected.
- Floating-point noise accepted by the previous implementation remains accepted.
- Extra off-diagonal entries are rejected.
- Matrices with too few rows are rejected.
- Mutations made through the zero-copy `to_sparse()` result are revalidated on every subsequent reconciliation.

## Validation

- `uv run pytest tests/test_core.py tests/test_utils.py --no-cov -q`
- `uv run ruff check hierarchicalforecast/core.py hierarchicalforecast/utils.py tests/test_core.py tests/test_utils.py`

### DIFF
--- a/hierarchicalforecast/core.py
+++ b/hierarchicalforecast/core.py
@@ -4,7 +4,6 @@ __all__ = ['HierarchicalReconciliation']
 import re
 import reprlib
 import time
-import warnings
 from inspect import signature
 
 import narwhals.stable.v2 as nw
@@ -256,7 +255,6 @@ class HierarchicalReconciliation:
         target_col: str = "y",
         id_time_col: str = "temporal_id",
         temporal: bool = False,
-        skip_identity_check: bool = False,
     ) -> tuple[FrameT, FrameT, FrameT, list[str], str]:
         """Performs preliminary wrangling and protections."""
         Y_hat_nw_cols = Y_hat_nw.columns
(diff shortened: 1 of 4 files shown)


hierarchicalforecast/core.py
elephaint, line 1: Thanks - Maybe we can do this a bit simpler; the issue seems to be that the check is slow, so, we can also just make the check fast. That would be simpler and require no API changes.

elephaint: This is where we'd have to change the code a bit. Snippet (see mainly the else branch, that avoids an allocation, which will reduce most of the cost)

```
# Assert S is an identity matrix at the bottom
S_nw_cols.remove(id_col)
if not skip_identity_check:
    n = len(S_nw_cols)
    # Slice rows first: only an n x n block is materialised (was m x n).
    S_bottom_nw = S_nw[-n:][S_nw_cols]
    S_bottom = S_bottom_nw.to_native()
    is_sparse_df = (
        hasattr(S_bottom, "sparse")
        and hasattr(S_bottom, "dtypes")
        and all(str(dtype).startswith("Sparse") for dtype in S_bottom.dtypes)
    )
    if is_sparse_df:
        ...  # unchanged
    else:
        B = S_bottom_nw.to_numpy()
        is_identity = (
            B.shape[0] == B.shape[1]
            and np.count_nonzero(B) == n
            and np.all(np.diagonal(B) == 1.0)
        )
        if not is_identity:
            raise ValueError(
                f"The bottom {n}x{n} part of S must be an identity matrix."
            )
```

elephaint (changes requested):
Thanks! I think we could do this a bit different, the remedy imho needs to fix the issue (the check is slow) rather than skipping it, which also potentially introduces new issues.

Apart from the code snippet I provided below, it would be worthwhile to address the repeat-call case properly, which I think would be a nice addition! E.g. add a boolean to SMatrix.__init__ as a cache so that the check doesn't get repeated or have `aggregate` return an SMatrix with a "verified bottom identity" (ie, provenance of the identity in the construction of the SMatrix).

Wdyt?

janrth:
> Thanks! I think we could do this a bit different, the remedy imho needs to fix the issue (the check is slow) rather than skipping it, which also potentially introduces new issues.
> 
> Apart from the code snippet I provided below, it would be worthwhile to address the repeat-call case properly, which I think would be a nice addition! E.g. add a boolean to SMatrix.**init** as a cache so that the check doesn't get repeated or have `aggregate` return an SMatrix with a "verified bottom identity" (ie, provenance of the identity in the construction of the SMatrix).
> 
> Wdyt?

I guess you are right. This is a lot of information for me. I need to let that sink in and will come back with a proposal (I hope).

hierarchicalforecast/core.py
janrth: Let's see what you are saying.

janrth:
Thanks for the feedback. The new and faster check is as fast as when I skipped the check. But the constraint here is the number of low level ids, I guess. But for most applications, this should be fine now in terms of speed. But please, have a close review. The underlying mechanics of hier. rec. is not my strength. So I have to trust codex here a lot.

elephaint (approved):

PR #494, opened 2026-08-03
perf: vectorize bootstrap sampling and sparse child-node lookup

loama:
## What

Three pure-Python/numpy hotspots, no changes to the C++ extension in `src/`:

1. `Bootstrap.get_samples` (`hierarchicalforecast/probabilistic_methods.py`) reconciled every bootstrap sample path with `np.apply_along_axis`, which is a hidden Python loop calling a lambda once per `(sample, horizon)` pair. Replaced with a single batched matmul.
2. `PERMBU._obtain_ranks` (same file) filled a ranks matrix with an explicit Python `for` loop over rows. Replaced with a single `np.put_along_axis` call.
3. `_get_child_nodes` (`hierarchicalforecast/methods.py`), which backs `TopDown`/`MiddleOut` `forecast_proportions`, unconditionally densified `S` with `S.toarray()` and then filtered each parent's children with a Python `idx in idx_node` membership scan over a numpy array (worst case `O(n_parents * n_bottom**2)`). Replaced with a sparse boolean matmul plus O(1) column lookups, and `S` is never densified.

`PERMBU`'s per-level `random_permutation` loop (the other spot called out in scope) was investigated and deliberately left untouched — see "The RNG decision" below.

## Why

All three were flagged as pure-Python-loop-shaped hotspots with no numerical dependency on the C++ backend. `Bootstrap.get_samples` and `_get_child_nodes` scale badly (the former re-enters Python once per sample/horizon pair, the latter is quadratic in the bottom-level series count), which matters for larger hierarchies (thousands of bottom series, hundreds/thousands of bootstrap samples).

### 1. `Bootstrap.get_samples`

Before:
```python
samples = [self.y_hat + residuals[:, idx:(idx+h)] for idx in samples_idx]
SP = self.S @ self.P
samples = np.apply_along_axis(lambda path: SP @ path, axis=1, arr=samples)
```

After, the sampled paths are stacked into `(num_samples, n_series, h)`, then reshaped so a single dense matmul reconciles every sample and horizon at once:

```python
samples_arr = np.stack(samples)
n_samples, n_series, _ = samples_arr.shape  # last dim is `h`; reuse the outer `h`, don't rebind
flat = np.moveaxis(samples_arr, 1, 0).reshape(n_series, n_samples * h)
del samples, samples_arr  # both are independent copies of the sampled paths by this point
reconciled = SP @ flat
samples_np = reconciled.reshape(SP.shape[0], n_samples, h)
```

Note: an `np.einsum('ij,sjh->sih', SP, samples_arr)` version was tried first and is numerically identical, but **`np.einsum` without `optimize=True` does not route this contraction through BLAS** — it fell back to a naive summation loop and was ~60x *slower* than the original `apply_along_axis` code on the benchmark hierarchy (12.8s vs the reshape+matmul's 0.2s in an isolated microbenchmark). The reshape+`@` form always uses a single `dgemm` call and doesn't depend on numpy's einsum path-optimizer, so it's the one shipped here.

**`SP` stays sparse when `S`/`P` are sparse** — an earlier version of this diff called `SP.toarray()` on the theory that `SP` (`base x base`) is "much smaller than `S`", which is wrong: `SP` is `base x base`, which is `>= S`'s `base x bottom`, so densifying it would OOM the `*Sparse` reconcilers on large hierarchies (e.g. ~8TB at 1M series). `SP @ flat` and `SP.shape[0]` both work directly against a sparse `SP`, so the `.toarray()` call was removed; a reviewer's sparse-vs-dense probe confirmed the two paths agree to 1e-12.

**Peak memory**: `samples` and `samples_arr` are dropped (`del`) right after the `flat` reshape/copy exists, since both are redundant copies of the same data by that point. Isolating just this one-line change (reshape+matmul with vs. without the `del`, same algorithm both sides, at `n_series=1021`, `h=12`, `num_samples=20000`, macOS `/usr/bin/time -l` peak RSS): **~6.1GB with the `del` vs. ~8.1GB without** (~1.3x lower peak) — real, reproduced numbers from this machine, not the original estimate. Comparing the full old (`apply_along_axis`) implementation against the new one at the same shape, wall time dropped from **~35-47s to ~3.0-3.3s (~11-14x)**; the peak-RSS delta between old and new algorithms was harder to isolate cleanly at the process level (both hover near ~6.1GB, since the ~2GB output array itself dominates the measurement and macOS's allocator doesn't always return freed pages to the OS immediately), but the `del`-only comparison above shows the memory saving is real.

Verified with `np.testing.assert_allclose` against a reimplementation of the old `apply_along_axis` logic (dense and sparse `S`/`P`); max abs diff was `~3e-10`, consistent with float summation-order differences.

### 2. `PERMBU._obtain_ranks`

```python
temp = array.argsort(axis=1)
ranks = np.empty_like(temp)
a_range = np.broadcast_to(np.arange(temp.shape[1]), temp.shape)
np.put_along_axis(ranks, temp, a_range, axis=1)
```

Verified with `np.testing.assert_array_equal` (exact integer equality, including a tie-breaking case) against the old loop. Honesty check on performance: at realistic sizes (hundreds to ~100k rows) this is **not a measurable win** — the old loop's per-row body was already a fully vectorized numpy assignment, so removing the Python `for` only saves loop overhead that was already dominated by the row-sized work inside it. Measured speedup was ~0.8x–1.5x (noise-level) across row counts from 100 to 100,000. Kept anyway because it's exactly equivalent, removes an explicit loop, and is what was asked for — but it is not being sold as a performance fix.

### 3. `_get_child_nodes`

Before, for every `(level, parent)` pair: `S.toarray()` once up front, then `child = np.zeros_like(S); child[idx_child] = S[idx_child]`, then per parent `idx_node = child * parent_node.astype(bool)` (dense elementwise multiply over the *entire* `S`) followed by `[idx for idx in idx_child if idx in idx_node]`, a Python-level linear scan.

After: for each level pair, extract only the `parent`/`idx_child` rows as boolean sparse matrices (never densifying all of `S`), then a single sparse matmul `child_bool @ parent_bool.T` gives, in one shot, which child rows overlap with which parent rows (a shared nonzero bottom-level column — the same condition the old dense code checked). Each parent's children are then read off via `O(1)` CSC column slicing (`indptr`/`indices`), with `sort_indices()` guaranteeing they come out in the same relative order as the original `idx_child` array (this was the subtle part — the old code implicitly preserved `idx_child`'s order via the list comprehension, and the new code has to replicate that explicitly rather than relying on `np.where`'s ascending order).

Verified against a reimplementation of the old dense/`toarray()` logic for: the existing 3-level fixture, sparse and dense `S` inputs, and randomly generated multi-level (up to 4 levels) hierarchies with **shuffled (non-ascending) tag ordering** — specifically to catch a rewrite that accidentally returns children sorted by row index instead of by `idx_child`'s given order.

### The RNG decision

`PERMBU.get_samples` also has this loop, mentioned in scope:

```python
random_permutation = np.array([
    rng.permutation(np.arange(num_samples)) for serie in range(len(parent_samples))
])
```

This is replaceable with `rng.permuted(...)`, but doing so **changes the RNG draw sequence** — same seed, different sample stream. Checked the test suite (`tests/test_probabilistic_methods.py`, `tests/test_core.py`) for anything pinning exact seeded PERMBU output: there is none (existing seed tests for PERMBU/Bootstrap only check reproducibility — same seed in, same output out — and shape, not literal values).

Even so, benchmarked both versions (Python-loop-of-`rng.permutation` vs one `rng.permuted` call) across realistic `(n_parents, num_samples)` combinations for this loop specifically (n_parents is the node count at one hierarchy level, typically tens, not thousands):

| n_parents | num_samples | old (loop) | new (`permuted`) | speedup |
|---:|---:|---:|---:|---:|
| 5 | 1000 | 2.5ms | 2.2ms | 1.15x |
| 20 | 1000 | 11.8ms | 8.0ms | 1.47x |
| 100 | 100 | 13.8ms | 5.0ms | 2.73x |
| 100 | 5000 | 333.8ms | 288.0ms | 1.16x |

The gain tops out around ~2.7x only at small `num_samples`, and is closer to 1.1–1.3x at realistic sizes, because the permutation algorithm itself (Fisher–Yates, `O(num_samples)`) dominates the cost either way — the Python loop overhead being removed was never the bottleneck here (unlike `_get_child_nodes`, where the removed loop was the asymptotic bottleneck). Per the "default to the conservative option for RNG-affecting lines unless the speedup is meaningful" rule: no test pins exact values, but the speedup isn't meaningful enough to justify a public, silent change to every downstream user's seeded PERMBU sample stream. **Left untouched.** Happy to revisit if maintainers consider a documented RNG-stream-breaking change acceptable for this magnitude of gain.

## Benchmarks

Synthetic hierarchy: 1000 bottom series, 3 levels (top=1, mid=20, bottom=1000), `num_samples=1000` where applicable. Measured by reimplementing the old logic inline and timing both against the same inputs (`min` of 3–10 repeats after warmup); see method in the PR — not committed as a script. Machine: Apple Silicon (arm64) laptop.

| Function | Old | New | Speedup |
|---|---:|---:|---:|
| `Bootstrap.get_samples` | ~2.88s | ~0.18s | **~16x** |
| `_get_child_nodes` (dense `S`) | ~133ms | ~7.6ms | **~17x** |
| `_get_child_nodes` (sparse `S`) | ~154ms | ~0.6ms | **~274x** |
| `PERMBU._obtain_ranks` | ~22ms | ~23ms | ~1x (no real change, see above) |

`_get_child_nodes` scaling with bottom-series count (dense vs dense, sparse vs sparse, same shape both sides):

| n_bottom | old (dense) | new (dense) | speedup | old (sparse) | new (sparse) | speedup |
|---:|---:|---:|---:|---:|---:|---:|
| 200 | 11.2ms | 0.98ms | 11.4x | 7.4ms | 0.36ms | 20.8x |
| 1000 | 133.2ms | 7.6ms | 17.5x | 153.7ms | 0.56ms | 273.6x |
| 5000 | 1413.2ms | 141.1ms | 10.0x | 1555.6ms | 1.0ms | 1562x |

The sparse-`S` speedup grows with hierarchy size because the old code paid the same `O(n_bottom^2)` cost regardless of input sparsity (it densified first), while the new code's cost tracks the actual number of nonzeros.

`pytest-benchmark` fixtures for `Bootstrap.get_samples` and `_get_child_nodes` were added to `tests/test_benchmarks.py` (same synthetic hierarchy), each parametrized over dense and sparse `S`/`P` so the sparse code paths get benchmark coverage too, not just the dense ones; they measure the current (post-fix) implementation only, since `pytest-benchmark` isn't set up here to diff against a prior git revision.

`tests/test_benchmarks.py` is excluded from the cibuildwheel CI test matrix (`.github/workflows/pytest.yml`): every `test_bench_*` function is now tagged `benchmark` (registered in `pyproject.toml`), and `CIBW_TEST_COMMAND` passes `-m "not benchmark"`. Without this, the ~20s of benchmark timing runs would execute in all 20 matrix jobs, including the QEMU-emulated `manylinux_aarch64` job, where they could balloon well past their normal duration. Benchmarks still run in a plain `pytest` invocation or explicitly via `-m benchmark`.

## Post-review fixes

This branch went through a round of code review; all findings were addressed:

- **Blocking**: removed the `SP.toarray()` densification in `Bootstrap.get_samples` (see the sparse-`SP` note above); added the `del samples, samples_arr` peak-memory fix (see above); excluded `tests/test_benchmarks.py` from the wheel test matrix (see above).
- **Minors**: replaced the per-process-salted `hash()` fuzz seed in `test_methods.py` with a deterministic `zlib.crc32` seed; reworded stale "einsum" docstrings/test names in `test_probabilistic_methods.py` to "batched reshape + matmul"; removed dead-code `nodes_level` dict-comprehension initializer in `_get_child_nodes` (immediately overwritten by the loop); documented the nonnegative-`S` assumption behind `_get_child_nodes`'s shared-nonzero-column overlap test and the sparse-`S`/`P` behavior of `Bootstrap.get_samples`, both as docstring notes; added a `num_samples=1`/`h=1` Bootstrap regression test and a hand-written singleton-child-parent `_get_child_nodes` fixture (random fuzzing isn't guaranteed to produce one); moved the duplicated `_make_random_strict_hierarchy` test helper (drifted between `test_methods.py` and `test_benchmarks.py`) into `conftest.py` for both to share; stopped rebinding `h` to a same-valued shadowing local in `Bootstrap.get_samples`.
- **Housekeeping**: fixes were folded into their original commits by concern rather than appended as a fixup — `perf(probabilistic): vectorize Bootstrap.get_samples reconciliation`, `perf(probabilistic): vectorize PERMBU rank computation`, `perf(methods): avoid densifying S in _get_child_nodes`, `test: add regression tests and benchmarks for perf fixes`, plus a new `ci: exclude benchmarks from wheel test matrix` commit for the CI-only change.

## Testing

- `git submodule update --init --recursive`, then `uv sync` (built the pybind11/Eigen C++ extension successfully; see note below).
- Ran the specific affected test files:
  - `tests/test_probabilistic_methods.py` — **53 passed** (includes the new `TestBootstrapVectorization`/`TestPermbuObtainRanksVectorization` cases plus the post-review `num_samples=1`/`h=1` edge case)
  - `tests/test_methods.py` — **133 passed** (includes `TestGetChildNodes`, plus the post-review singleton-child-parent case, plus the deterministic-seed fuzz test)
  - `tests/test_benchmarks.py` — **17 passed** with `-m benchmark --benchmark-disable` (functional correctness, not timing); `-m benchmark --collect-only` collects all 17, `-m "not benchmark" --collect-only` collects 0 from this file, confirming the CI marker filter works as intended.
- Added regression tests reimplementing the pre-optimization logic inline and asserting equivalence:
  - `TestBootstrapVectorization` — `np.testing.assert_allclose` (dense and sparse `S`/`P`, plus a dedicated degenerate `num_samples=1`/`h=1` case)
  - `TestPermbuObtainRanksVectorization` — `np.testing.assert_array_equal` (exact), including a tie-breaking case, parametrized over several shapes
  - `TestGetChildNodes` — exact structural/order equality against the reference dense implementation, including a fuzz test over random multi-level hierarchies with shuffled tag ordering (both dense and sparse `S`, now with a deterministic `crc32`-derived seed instead of `hash()`), and a hand-written hierarchy that forces a singleton-child parent
- Full-suite run in this sandbox (`tests/`, excluding benchmarks) hits network-dependent fixtures (`tourism_df` et al. fetch a CSV from `raw.githubusercontent.com`, which this particular sandbox can't reach) that error out unrelated to this change — same failures exist on `main`. The files this PR actually touches (`test_probabilistic_methods.py`, `test_methods.py`, `test_benchmarks.py`) have no such network dependency and all pass cleanly as reported above.
- `ruff check` passes clean on every file touched by this PR, including the new `conftest.py` additions and the CI/`pyproject.toml` changes. `ruff format --check`: `test_benchmarks.py` is kept format-clean (it already was on `main`, and stayed that way through the post-review edits — line-wrapped the one call that got too long); the other four touched Python files (`probabilistic_methods.py`, `methods.py`, `test_methods.py`, `test_probabilistic_methods.py`, `conftest.py`) were already not `ruff format`-clean on `main` before this branch, and this PR doesn't newly introduce that — left as-is rather than mixing an unrelated repo-wide reformat into a perf/test PR.

### Build note (environment-specific, not a code issue)

Building the C++ extension locally required working around a local toolchain quirk: this machine is arm64 but its only Homebrew install is the x86_64-prefix one, so `brew --prefix libomp` resolved to an x86_64-only `libomp.dylib`/`omp.h`, and the resulting universal2 build's arm64 slice silently linked with no OpenMP symbols at all, then failed at import with `symbol not found in flat namespace '__kmpc_dispatch_deinit'` once an arm64 `libomp` was substituted from an older bundled copy. Fixed locally by pulling the current arm64 `libomp` Homebrew bottle directly and building `ARCHFLAGS="-arch arm64"`-only. This is purely a local dev-environment issue on this particular machine (mismatched Homebrew prefix vs. host architecture) and isn't something this PR needs to address in `setup.py`.

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -58,4 +58,4 @@ jobs:
             uv pip install datasetsforecast "statsforecast>=1.0.0"
             scipy "polars[numpy]<=1.32.0" pytest pytest-benchmark pytest-cov
           CIBW_TEST_COMMAND: >
-            pytest {project}/tests --no-cov --import-mode=importlib
+            pytest {project}/tests --no-cov --import-mode=importlib -m "not benchmark"
--- a/hierarchicalforecast/methods.py
+++ b/hierarchicalforecast/methods.py
@@ -340,24 +340,54 @@ class BottomUpSparse(BottomUp):
         return P, W
 
 
+def _rows_to_bool_csr(S: np.ndarray | sparse.spmatrix, rows: np.ndarray) -> sparse.csr_matrix:
+    """Extract `rows` from `S` as a boolean sparse matrix.
+
+    Only the requested rows are ever materialized, so this never densifies a
+    (potentially huge) sparse `S`, unlike a blanket `S.toarray()`.
+    """
+    if sparse.issparse(S):
+        sub = S.tocsr()[rows].astype(bool)
+        sub.eliminate_zeros()
+        return sub
+    return sparse.csr_matrix(np.asarray(S)[rows] != 0)
+
+
 def _get_child_nodes(
     S: np.ndarray | sparse.csr_matrix, tags: dict[str, np.ndarray]
 ):
-    if isinstance(S, sparse.spmatrix):
-        S = S.toarray()
+    """Build, for each level, a mapping from parent index to its child indices.
+
+    Note:
+        The child-parent overlap test assumes `S` is nonnegative (true for
+        standard 0/1 hierarchical summing matrices): it checks for any shared
+        nonzero bottom-level column between a child row and a parent row,
+        which only matches "sum of overlapping values > 0" when there is no
+        sign cancellation between entries.
+    """
     level_names = list(tags.keys())
     nodes = OrderedDict()
     for i_level, level in enumerate(level_names[:-1]):
         parent = tags[level]
-        child = np.zeros_like(S)
         idx_child = tags[level_names[i_level + 1]]
-        child[idx_child] = S[idx_child]
+
+        parent_bool = _rows_to_bool_csr(S, parent)
+        child_bool = _rows_to_bool_csr(S, idx_child)
+
+        # overlap[c, p] is nonzero iff child row `c` and parent row `p` share at
+        # least one nonzero bottom-level column (i.e. child `c` descends from
+        # parent `p`). This replaces the dense `child * parent_node.astype(bool)`
+        # check (and the O(n_bottom^2)-worst-case `idx in idx_node` membership
+        # scan) with a single sparse matmul plus O(1) column lookups.
+        overlap = (child_bool @ parent_bool.T).tocsc()
+        overlap.sort_indices()
+
         nodes_level = {}
-        for idx_parent_node in parent:
-            parent_node = S[idx_parent_node]
-            idx_node = child * parent_node.astype(bool)
-            (idx_node,) = np.where(idx_node.sum(axis=1) > 0)
-            nodes_level[idx_parent_node] = [idx for idx in idx_child if idx in idx_node]
+        for p_pos, idx_parent_node in enumerate(parent):
+            start, end = overlap.indptr[p_pos], overlap.indptr[p_pos + 1]
+            nodes_level[idx_parent_node] = [
+                idx_child[c_pos] for c_pos in overlap.indices[start:end]
+            ]
         nodes[level] = nodes_level
     return nodes
 
--- a/hierarchicalforecast/probabilistic_methods.py
+++ b/hierarchicalforecast/probabilistic_methods.py
@@ -570,6 +570,12 @@ class Bootstrap:
 
         Returns:
             samples: Coherent samples of size (`base`, `horizon`, `num_samples`).
+
+        Note:
+            When `S` and/or `P` are sparse, `SP = S @ P` stays sparse and the
+            reconciliation matmul is a sparse-dense product; `SP` is never
+            densified, so memory stays proportional to `S`/`P`'s nonzeros
+            instead of `base x base`.
         """
         residuals = self.y_insample - self.y_hat_insample
         h = self.y_hat.shape[1]
@@ -581,11 +587,34 @@ class Bootstrap:
         samples_idx = rng.choice(sample_idx, size=num_samples)
         samples = [self.y_hat + residuals[:, idx : (idx + h)] for idx in samples_idx]
         SP = self.S @ self.P
-        samples = np.apply_along_axis(lambda path: SP @ path, axis=1, arr=samples)
-        samples_np = np.stack(samples)
 
-        # [samples, N, H] -> [N, H, samples]
-        samples_np = samples_np.transpose((1, 2, 0))
+        # Stack into a single (num_samples, n_series, h) array and reconcile all
+        # samples with one batched matmul instead of looping via
+        # `np.apply_along_axis` (which calls the lambda once per (sample, horizon)
+        # pair under the hood). Flattening (samples, horizon) into one axis lets
+        # this go through a single BLAS `dgemm` call (or scipy sparse matmul,
+        # when `SP` is sparse, which is left sparse here so large hierarchies
+        # don't force an OOM-inducing dense `SP` allocation) rather than
+        # `np.einsum`, which is dramatically slower here since it can't
+        # express this contraction as a plain matrix product without
+        # `optimize=True`.
+        samples_arr = np.stack(samples)
+        # samples_arr.shape[-1] is `h` by construction (each entry is
+        # `y_hat + residuals[:, idx:(idx+h)]`), so we reuse the outer `h`
+        # instead of rebinding it to a shadowing local of the same value.
+        n_samples, n_series, _ = samples_arr.shape
+        flat = np.moveaxis(samples_arr, 1, 0).reshape(n_series, n_samples * h)
+        # `samples` and `samples_arr` are independent copies of the sampled
+        # paths at this point (the moveaxis+reshape above produced its own
+        # copy in `flat`), so drop them before the matmul to keep peak memory
+        # to roughly 2x the output buffer instead of holding every
+        # intermediate alive at once.
+        del samples, samples_arr
+        reconciled = SP @ flat
+        samples_np = reconciled.reshape(SP.shape[0], n_samples, h)
+
+        # [N, samples, H] -> [N, H, samples]
+        samples_np = samples_np.transpose((0, 2, 1))
         return samples_np
 
     def get_prediction_levels(self, res, level):
--- a/hierarchicalforecast/probabilistic_methods.py
+++ b/hierarchicalforecast/probabilistic_methods.py
@@ -713,9 +713,8 @@ class PERMBU:
         """
         temp = array.argsort(axis=1)
         ranks = np.empty_like(temp)
-        a_range = np.arange(temp.shape[1])
-        for i_row in range(temp.shape[0]):
-            ranks[i_row, temp[i_row, :]] = a_range
+        a_range = np.broadcast_to(np.arange(temp.shape[1]), temp.shape)
+        np.put_along_axis(ranks, temp, a_range, axis=1)
         return ranks
 
     def _permutate_samples(self, samples, permutations):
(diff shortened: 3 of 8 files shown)


loama:

PR #495, opened 2026-08-04
perf: matrix-free kronecker lasso for erm reconciliation

loama:
## What

`ERM(method="reg")` and `ERM(method="reg_bu")` no longer materialize the Lasso design matrix `X = np.kron(S, y_hat_insample.T)`. A new C++ kernel (`_lasso_kron`, alongside the existing `_lasso` in `src/reconciliation.cpp`) runs the same cyclic coordinate descent directly on the Kronecker factors `S` (`n_hiers x n_bottom`) and `Y = y_hat_insample.T` (`h x n_hiers`).

### Why: the memory math

`X` has shape `(n_hiers * h, n_bottom * n_hiers)`:

- **Old:** `8 * n_hiers^2 * n_bottom * h` bytes for `X`, and the `_lasso` wrapper's `np.asfortranarray(X)` copies it again, so peak is ~**2x** that.
- **New:** `O(n_hiers * n_bottom + n_hiers * h)` — just the factors plus a residual vector.

Concrete sizes at `h = 12`:

| Hierarchy | `X` alone | Old peak (`X` + F-order copy) | New extra memory |
| --- | --- | --- | --- |
| Australian tourism, grouped (425 series, 304 bottom) | 4.9 GiB | ~9.8 GiB | ~1 MiB |
| 2,000 series, 1,500 bottom | 536 GiB | ~1.05 TiB | ~24 MiB |

The quadratic-in-`n_hiers` term is gone entirely. (The existing cap of insample steps at `h` "to avoid computational burden" is preserved unchanged, so results are identical; with the new kernel that cap could arguably be relaxed in a follow-up.)

## How

`np.kron` layout: `X[i1*h + i2, j1*n_hiers + j2] = S[i1, j1] * Y[i2, j2]`. Grouping the column index as `j = j1 * n_hiers + j2` gives

```
X[:, j] = np.kron(S[:, j1], Y[:, j2])
```

so every quantity the coordinate-descent kernel needs factors through `S` and `Y`:

- column norms: `||X[:, j]||^2 = ||S[:, j1]||^2 * ||Y[:, j2]||^2`
- residual dot: `X[:, j] . r = sum_i1 S[i1, j1] * (Y[:, j2] . r[i1*h : (i1+1)*h])`
- update: `r += a * X[:, j]`  ->  `r[i1*h : (i1+1)*h] += a * S[i1, j1] * Y[:, j2]`

Zero entries of `S[:, j1]` contribute exactly zero to these sums and are skipped; summing-matrix columns have only `O(depth)` nonzeros, so each coordinate update also drops from `O(n_hiers * h)` to `O(depth * h)` flops on top of the memory win.

The same identities remove the remaining `X` uses in `ERM._get_PW_matrices`:

- `reg_bu`'s `X @ Pbu.reshape(-1)` becomes `S @ Pbu.reshape(n_bottom, n_hiers) @ y_hat_insample` (the flat index regrouped as `V[j1, j2] = Pbu.flat[j1*n_hiers + j2]`),
- `lambda_reg=None`'s `max |X.T @ z|` becomes `max |S.T @ Z @ Y|` with `Z = z.reshape(n_hiers, h)`.

`ERM._get_PW_matrices` now also accepts a sparse `S`, densified up front (`S` is the small factor). `np.kron` itself does not raise on a sparse `S` — it silently returns an object-dtype array of sparse-matrix blocks — so the previous code failed downstream instead: `TypeError` in `_lasso` (pybind11 rejects the object dtype), `ValueError` in the `lambda_reg=None` bound for `reg`/`reg_bu` (`X.T @ z` on the object array), and `LinAlgError` in `np.linalg.inv` for `closed`. Densifying `S` up front fixes all three at the source, a strict capability improvement rather than a workaround for one specific failure mode.

### Equivalence strategy

`_lasso_kron` mirrors `_lasso` exactly: same coordinate iteration order (`j = j1 * n_hiers + j2` ascending), same soft-threshold formula, same `norms < 1e-8` skip guard, same `max|delta beta| < tol` convergence criterion, same `max_iters`/`tol` defaults. The only difference is that the reductions are computed through the Kronecker factors, i.e. identical arithmetic reassociated. **Achieved equivalence level: floating-point round-off (allclose), not bit-equality** — reassociating the reductions necessarily changes rounding order. Measured against the materialized `np.kron` + `_lasso` path:

- kernel fuzzing over shapes (including `n_bottom=1`, single-level, `h=1`): max abs deviation ~2e-14
- full `_get_PW_matrices` at 181 series x 150 bottom, h=12: max abs deviation ~3e-13

Tests assert `rtol=1e-8, atol=1e-11`.

## Benchmarks (Apple M-series, single thread)

`pytest-benchmark`, `tests/test_benchmarks.py`, implicit `X` of 448 x 2,800 (56-series hierarchy, 50 bottom, h=8):

| Benchmark | Median | Speedup |
| --- | --- | --- |
| `test_bench_lasso_kron_materialized_baseline` (old: `np.kron` + `_lasso`) | 2,446 us | 1x |
| `test_bench_lasso_kron` (new, matrix-free) | 37.1 us | **66x** |

End-to-end `ERM._get_PW_matrices` (both Lasso methods), peak RSS via `ru_maxrss`, each path in a fresh process:

**Australian tourism grouped hierarchy (425 series, 304 bottom, h=12)** — `X` is 5,100 x 129,200 = 4.91 GiB:

| Path | reg | reg_bu | Peak RSS |
| --- | --- | --- | --- |
| Old (`np.kron` + `_lasso`) | 484.9 s | 315.3 s | 9.97 GiB |
| New (`_lasso_kron`) | 9.9 s (**49x**) | 8.8 s (**36x**) | **0.29 GiB** (34x less) |

**Mid-scale synthetic (181 series, 150 bottom, h=12)** — `X` is 2,172 x 27,150 = 0.44 GiB:

| Path | reg | reg_bu |
| --- | --- | --- |
| Old | 30.8 s | 31.8 s |
| New | 0.9 s (**34x**) | 1.5 s (**21x**) |

**Capability unlock — 2,000-series hierarchy (1,500 bottom, 3 aggregation levels, h=12):** the old path must allocate a 24,000 x 3,000,000 design matrix: **536 GiB for `X` plus an equal-size F-order copy (~1.05 TiB peak)**, far beyond the 64 GiB test machine — it cannot run at all. The new path:

| Path | reg | reg_bu | Peak RSS |
| --- | --- | --- | --- |
| New (`_lasso_kron`) | 58.5 s | 92.9 s | **0.41 GiB** |

## Testing

- `tests/test_methods.py`:
  - `test_lasso_kron_matches_materialized_kron` — parametrized over edge shapes (`n_bottom=1`, `h=1`, single-level, grouped, deep) x two lambdas, asserts the matrix-free kernel matches `_lasso(np.kron(S, Y), ...)` at `rtol=1e-8, atol=1e-11`.
  - `test_lasso_kron_fuzz_random_shapes` — 25 randomized shape/lambda configs against the materialized path.
  - `test_lasso_kron_validates_target_length` — input validation on `y`'s length.
  - `test_lasso_kron_validates_kronecker_shapes` — input validation on `S.rows() == Y.cols()`.
  - `test_lasso_kron_zeroed_column_matches_materialized_kron` — a zeroed `Y` column drives `norms_i` below the `1e-8` skip guard for every `S` column at that coordinate; still matches the materialized path.
  - `test_erm_reg_matches_materialized_kron` — `reg`/`reg_bu` x `lambda_reg in {1e-2, None}`: `ERM._get_PW_matrices` vs an inline re-implementation of the old materialized path.
  - `test_erm_accepts_sparse_S` — `closed`/`reg`/`reg_bu` with a `csr_matrix` summing matrix match the dense results.
- `tests/test_core.py`: `test_erm_lasso_methods_work_through_reconcile` — `reg`/`reg_bu` end to end through the public `HierarchicalReconciliation.reconcile` API, parametrized over both the tourism strict hierarchy and the full grouped hierarchy (425 series, 304 bottom) — previously only `closed` was tested on the grouped hierarchy, and `reg`/`reg_bu` weren't tested through `reconcile` at all, because the materialized path was too expensive.
- `tests/test_benchmarks.py`: `test_bench_lasso_kron` + `test_bench_lasso_kron_materialized_baseline`, following the file's existing fixture/bench patterns (no markers, consistent with main's current state). Note: #494 also touches `tests/test_benchmarks.py`, so whichever lands second has a small, mechanical merge overlap in this file.
- Full suites green locally: `tests/test_methods.py` (137), `tests/test_core.py` (76), `tests/test_utils.py` (20), `tests/test_probabilistic_methods.py` + `tests/test_evaluation.py` + `tests/test_diagnostics.py` (65), `tests/test_benchmarks.py` lasso benches.

### DIFF
--- a/hierarchicalforecast/methods.py
+++ b/hierarchicalforecast/methods.py
@@ -14,7 +14,7 @@ from scipy import sparse
 from hierarchicalforecast.utils import (
     _construct_adjacency_matrix,
     _is_strictly_hierarchical,
-    _lasso,
+    _lasso_kron,
     _ma_cov,
     _shrunk_covariance_schaferstrimmer_no_nans,
     _shrunk_covariance_schaferstrimmer_with_nans,
@@ -2041,6 +2041,11 @@ class ERM(HReconciler):
         y_insample: np.ndarray,
         y_hat_insample: np.ndarray,
     ):
+        if sparse.issparse(S):
+            # S is the small Kronecker factor (n_hiers, n_bottom); every ERM
+            # variant needs it dense, so densify it here. The insample
+            # matrices dominate memory, not S.
+            S = S.toarray()
         n_hiers, n_bottom = S.shape
         # y_hat_insample shape (n_hiers, obs)
         if y_insample is None or y_hat_insample is None:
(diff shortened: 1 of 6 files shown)


loama:

PR #496, opened 2026-08-04
perf: native batched forecast-proportions traversal (stacked on #494)

loama:
> **STACKED PR — do not merge before #494.** This branch (`perf/native-forecast-proportions-traversal`) is built on top of `perf/vectorize-probabilistic-and-proportions` (PR #494) and reuses its vectorized `_get_child_nodes` and its test infrastructure (`_make_random_strict_hierarchy` in `tests/conftest.py`, the benchmark suite layout, and the CI benchmark exclusion). After #494 merges, rebase this branch onto `main` (or retarget the PR) — only the four commits ending at `test: add regression tests and benchmarks for native traversal` belong to this PR.

## What

`TopDown`/`MiddleOut` with `method="forecast_proportions"` reconciled forecasts with a 4-level nested pure-Python tree traversal (`_reconcile_fcst_proportions` in `hierarchicalforecast/methods.py`: top nodes x levels x parents x children, scalar numpy ops in the innermost loop). That traversal was re-executed:

- once per horizon column in `TopDown.fit_predict` (`for y_hat_ in y_hat.T`),
- once per **bootstrap sample x horizon column** for prediction intervals (`_reconcile_fcst_proportions_bootstrap`) — a 100–10,000x multiplier on the traversal count (`num_samples` defaults to 100 and is commonly 1,000+),
- and `MiddleOut` repeats all of the above once per cut node.

This PR moves the traversal into the existing pybind11/Eigen C++ extension as a batched kernel that processes **all (sample, horizon) pairs in one call**, and wires the dense `TopDown`/`MiddleOut` mean and bootstrap paths through it. The pure-Python implementation stays in `methods.py` as the behavioral reference for the regression tests. The sparse path (`_reconcile_fcst_proportions_sparse`, used by `TopDownSparse`/`MiddleOutSparse`) is untouched — it is already vectorized per column with sparse matmuls and would need sparse-matrix plumbing into C++; it can be a follow-up if it shows up in profiles.

## Why

For a hierarchy with `n` nodes and `h`-step forecasts, the old path performed `O(n)` Python-level scalar numpy operations per traversal and `h * num_samples` traversals per `fit_predict` with intervals. On the benchmark hierarchy below (1,021 nodes, `h=12`, 1,000 samples) that is 12,000 Python traversals ≈ 25M Python-level numpy scalar ops — 36 seconds of pure interpreter overhead for ~0.3s of actual arithmetic.

### Flattening (once per `fit_predict`)

`_flatten_child_nodes` converts the `_get_child_nodes` output (`{level: {parent: [children]}}`, already vectorized on #494) into three CSR-style int64 arrays: `parent_idx` (parents in traversal order: levels top-to-bottom, insertion order within a level), `child_indptr`, and `child_idx`. A single flat pass over that parent list is exactly equivalent to the reference's level-by-level traversal because every parent's reconciled value is written (as a top node, or as a child at the previous level) before it is read. Level offsets therefore don't need to be passed to the kernel at all. The same flattening also showed the reference's outer `for idx_top in idxs_top` loop is redundant re-computation: the final pass is identical to seeding all top nodes once and traversing once, which is what the kernel does.

### Kernel (`src/forecast_proportions.cpp`, new translation unit)

`_forecast_proportions_traversal(samples (S, n, h), parent_idx, child_indptr, child_idx, top_idx) -> (S, n, h)` follows the existing binding patterns in `src/reconciliation.cpp` (same submodule, `unchecked` buffers, GIL released around the compute region, input validation with `ValueError` on malformed/out-of-bounds flat arrays). The mean path calls it with `S=1`; the bootstrap path materializes all `num_samples` base-forecast matrices with one vectorized numpy gather+add and reconciles them in one call.

### Bit-exactness strategy

The goal was exact float equality with the pure-Python reference, not allclose:

- **Child sums**: the reference computes `y_hat[idx_childs].sum()` — NumPy's pairwise summation over a contiguous child-ordered vector. The kernel ports `DOUBLE_pairwise_sum` from `numpy/_core/src/umath/loops_utils.h.src` faithfully (sequential under 8 elements including the `-0.0` seed, the 8-accumulator unrolled block with its exact combination tree up to 128, recursive split rounded to a multiple of 8 above). NumPy ships that reduction as a single generic C source with no per-architecture SIMD dispatch variant — the unroll is explicitly designed so compilers can vectorize it *without changing summation order* — so the association tree is platform-independent. (Empirically necessary: NumPy's own `sum(axis=...)` over 2D/3D gathers diverges from the 1-D contiguous form in the last ulp for fan-outs ≥ 8, so precomputing the sums with batched numpy reductions was tried first and abandoned.)
- **Scalar chain**: the per-child update is the reference's `y * fcst_parent / childs_sum` (and `fcst_parent / n_children` under the `|sum| < 1e-8` guard) in the same operation order.
- **Fast-math containment (per-source build flags, not pragmas)**: the extension builds with `-ffast-math` (`/fp:fast` on MSVC), which would license reciprocal hoisting of the loop-invariant divisor (`x/s -> x * (1/s)`), reassociation, FP contraction — all last-ulp-changing — and, via finite-math-only, lets the compiler assume no NaN/inf. That last one is not a last-ulp issue but a correctness bug: with `-ffast-math` on clang the kernel's `|child_sum| < 1e-8` guard was allowed to assume the sum is never NaN, so a NaN child forecast (e.g. `y_hat = [10, 1, NaN, 2]` on a star hierarchy) silently took the equal-split branch and produced finite output (`[10, 3.33, 3.33, 3.33]`) where the Python reference propagates NaN (`[10, nan, nan, nan]`). In-file FP pragmas cannot fully undo a driver-level `-ffast-math` on every toolchain (clang's scoped pragmas in particular do not restore NaN semantics), so `setup.py` now excludes `src/forecast_proportions.cpp` from fast-math at the build level: a small `build_ext` override (`STRICT_FP_SOURCES`) appends per-source flags — `-fno-fast-math` on gcc/clang (the later flag wins over the earlier `-ffast-math`), `/fp:precise` on MSVC (same later-flag-wins rule over `/fp:fast`). The existing kernels' flags are untouched, and the in-file pragmas are kept as defense-in-depth only. NaN and inf + -inf child-sum propagation is pinned by regression tests, exact-equal against the reference.

### OpenMP parallelism boundaries

Parallelism spans **only the fully independent (sample, horizon) pairs** (flattened into one loop by hand — MSVC's OpenMP 2.0 has no `collapse`). No accumulation ever crosses a thread boundary and each pair writes disjoint output entries, so the result is deterministic and bit-identical to a serial run for any thread count (asserted: 1 thread vs 10 threads, `np.array_equal`).

## Benchmarks

Measured on this branch (arm64 macOS, 10 OpenMP threads, Python 3.12, numpy 2.3.5), `pytest tests/test_benchmarks.py -k fcst_proportions` — synthetic strict hierarchy with 1,000 bottom series, 3 levels (1,021 nodes), `h=12`:

| Workload | Python (before) | Native (after) | Speedup |
| --- | --- | --- | --- |
| Mean path (12 columns) | 45.3 ms | 0.311 ms | **~146x** |
| Bootstrap intervals, 100 samples | 3.62 s | 71.6 ms | **~51x** |
| Bootstrap intervals, 1,000 samples (one-off timing) | 36.12 s | 0.855 s | **~42x** |

The 1,000-sample outputs of the two implementations are bit-identical (`np.array_equal` on the returned quantiles). At 1,000 samples the native path's remaining time is dominated by `np.quantile` over the `(1021, 12, 1000)` sample tensor, not the traversal. The benchmark suite keeps the parametrized python-vs-native pairs at sizes that run in seconds (the 100-sample python case uses `benchmark.pedantic(rounds=3)`) plus a native-only 1,000-sample case; benchmarks remain excluded from the CI wheel matrix via `-m "not benchmark"` (from #494).

## Testing

- `TestForecastProportionsNative` in `tests/test_methods.py` (20 tests, all **exact equality** via `assert_array_equal`, no allclose):
  - fuzz vs the per-column Python reference over random strict hierarchies (`_make_random_strict_hierarchy`, crc32-derived seeds), covering fan-outs below 8 (sequential summation), above 8 (unrolled block), and a 200-child parent (recursive pairwise split above the 128 block size);
  - zero-children-sum branch (exact 0.0 and below-threshold 1e-10 sums), partial top-node seeding leaving unseeded subtrees untouched, batched-vs-single-sample consistency;
  - non-finite propagation: a NaN child and an inf + -inf child pair both poison the child sum and must propagate NaN exactly like the reference (the fast-math regression guard described above);
  - float32 fallback: non-float64 input matches the pre-native Python path bit-for-bit and preserves dtype end-to-end, on the direct call, the public `TopDown` mean path, and the bootstrap path;
  - end-to-end public API equivalence for `TopDown` and `MiddleOut` (fixture + random hierarchy) by monkeypatching `_reconcile_fcst_proportions_native` with a wrapper that routes every (sample, column) through the pure-Python `_reconcile_fcst_proportions`;
  - seeded bootstrap quantile equality (direct call and end-to-end through `TopDown(..., intervals_method="bootstrap")`, mean and quantiles);
  - kernel input validation (malformed `child_indptr` shape, `child_indptr[0] != 0`, non-decreasing violation, out-of-bounds parent/child/top indices all raise `ValueError`).
- Full suites green locally: `tests/test_methods.py` + `tests/test_probabilistic_methods.py` (206 passed), `tests/test_utils.py` (15 passed). (`tests/test_core.py`/`tests/test_evaluation.py` dataset-download fixtures fail in this sandboxed environment on the base branch already — network-restricted, unrelated.)
- Local extension build: compiles clean (arm64 clang, OpenMP via libomp); thread-count determinism asserted at 1 vs 10 threads.

## Notes for reviewers

- **Native for float64, Python reference for everything else** — the kernel computes in float64 only, and pybind11's `forcecast` would otherwise silently upcast e.g. float32 `y_hat` from direct `TopDown`/`MiddleOut` callers, changing both the arithmetic and the output dtype. `_reconcile_fcst_proportions_native` therefore gates on dtype: non-float64 input falls back to the retained pure-Python reference, reproducing the pre-native behavior exactly in the input's native dtype. `core.reconcile` coerces base forecasts to float64 upstream, so the main `reconcile()` path always takes the kernel — the gate only affects direct callers. (For float64, this matches how `_ma_cov`/`_lasso`/the Schafer–Strimmer kernels were ported: `_lib` is called unconditionally, no runtime fallback.)
- **Bootstrap memory scaling**: the dense bootstrap path materializes all base-forecast samples as one `(num_samples, n_nodes, h)` float64 tensor before the single kernel call. At the default 100 samples this is negligible, but at 10,000 samples on a large hierarchy (say 50k nodes, `h=12`) it is tens of GB — multi-GB territory starts around a few thousand samples on 10k+-node hierarchies. Chunking the sample axis (reconciling in fixed-size batches and streaming into the quantile buffer) is the natural follow-up if that becomes a real workload; the kernel already accepts arbitrary sample batch sizes, so no C++ changes would be needed.
- The `flat` tuple is threaded through `_reconcile_fcst_proportions_bootstrap` as an optional argument (computed once in `fit_predict`, alongside the existing `nodes`/`idxs_top`), so the public signature stays backward compatible.
- Cross-platform bit-exactness rests on NumPy's pairwise summation being the single generic C implementation described above (verified against numpy 2.3.5 sources and exhaustively size-swept locally; numpy >= 1.21 checked for the no-SIMD-variant claim, and numpy < 2.0's `0.` vs `-0.0` seed difference in the < 8 base case is unobservable through the `|sum| < 1e-8` guard). If a future NumPy release ever changes that reduction, the exact-equality fuzz tests will catch it loudly rather than silently drifting.
- The local-build libomp architecture caveat from #494 (an x86_64-prefix Homebrew on an arm64 machine resolving `brew --prefix libomp` to the wrong-arch dylib; fixed locally with the arm64 `libomp` bottle and an `ARCHFLAGS="-arch arm64"`-only build) applies to this branch's extension build too.

### DIFF
--- a/hierarchicalforecast/methods.py
+++ b/hierarchicalforecast/methods.py
@@ -11,6 +11,7 @@ import numpy as np
 from qpsolvers import solve_qp
 from scipy import sparse
 
+from hierarchicalforecast._lib import reconciliation as _lib_recon
 from hierarchicalforecast.utils import (
     _construct_adjacency_matrix,
     _is_strictly_hierarchical,
@@ -392,6 +393,93 @@ def _get_child_nodes(
     return nodes
 
 
+def _flatten_child_nodes(
+    nodes: dict[str, dict[int, np.ndarray]],
+) -> tuple[np.ndarray, np.ndarray, np.ndarray]:
+    """Flatten `_get_child_nodes` output into CSR-style int64 arrays.
+
+    Returns `(parent_idx, child_indptr, child_idx)` where parents appear in
+    traversal order (levels top-to-bottom, insertion order within each level)
+    and `child_idx[child_indptr[p]:child_indptr[p + 1]]` are the children of
+    node `parent_idx[p]`, preserving the exact ordering the pure-Python
+    reference `_reconcile_fcst_proportions` iterates in.
+    """
+    parent_idx: list[int] = []
+    child_indptr: list[int] = [0]
+    child_idx: list[int] = []
+    for nodes_level in nodes.values():
+        for idx_parent, idx_childs in nodes_level.items():
+            parent_idx.append(idx_parent)
+            child_idx.extend(idx_childs)
+            child_indptr.append(len(child_idx))
+    return (
+        np.asarray(parent_idx, dtype=np.int64),
+        np.asarray(child_indptr, dtype=np.int64),
+        np.asarray(child_idx, dtype=np.int64),
+    )
+
+
+def _reconcile_fcst_proportions_native(
+    y_hat_samples: np.ndarray,
+    idxs_top: np.ndarray,
+    flat: tuple[np.ndarray, np.ndarray, np.ndarray],
+) -> np.ndarray:
+    """Native batched forecast-proportions traversal.
+
+    Runs the `_reconcile_fcst_proportions` tree traversal for every
+    (sample, horizon) pair of `y_hat_samples` (`(n_samples, n_nodes, h)`) in
+    one C++ call, OpenMP-parallel across the independent pairs. The kernel
+    reproduces the reference's arithmetic exactly (NumPy's pairwise summation
+    for each parent's child sum, then the scalar multiply/divide chain), so
+    the output is bit-identical to running the pure-Python reference per
+    column.
+
+    The kernel computes in float64 only, so non-float64 input falls back to
+    the pure-Python reference, preserving the input dtype end-to-end.
+    """
+    parent_idx, child_indptr, child_idx = flat
+    if y_hat_samples.dtype != np.float64:
+        # The kernel accepts float64 only -- pybind11's `forcecast` would
+        # silently upcast e.g. float32 `y_hat` from direct TopDown/MiddleOut
+        # callers, changing both the arithmetic and the output dtype.
+        # `core.reconcile` coerces base forecasts to float64 upstream, so the
+        # main reconcile() path always takes the kernel; any other dtype runs
+        # the pure-Python reference below, which reproduces the pre-native
+        # behavior exactly in the input's native dtype. Presenting the flat
+        # parent list as a single level is equivalent to the reference's
+        # level-by-level traversal: parents are stored in traversal order, so
+        # every parent's reconciled value is written before it is read.
+        nodes_flat = {
+            "flat": {
+                int(p): child_idx[child_indptr[i] : child_indptr[i + 1]]
+                for i, p in enumerate(parent_idx)
+            }
+        }
+        tags_flat = {"flat": None, "leaves": None}
+        out = np.empty_like(y_hat_samples)
+        for s, y_hat_sample in enumerate(y_hat_samples):
+            out[s] = np.hstack(
+                [
+                    _reconcile_fcst_proportions(
+                        S=None,
+                        y_hat=y_hat_sample[:, [c]],
+                        tags=tags_flat,
+                        nodes=nodes_flat,
+                        idxs_top=idxs_top,
+                    )
+                    for c in range(y_hat_sample.shape[1])
+                ]
+            )
+        return out
+    return _lib_recon._forecast_proportions_traversal(
+        y_hat_samples,
+        parent_idx,
+        child_indptr,
+        child_idx,
+        np.ascontiguousarray(idxs_top, dtype=np.int64),
+    )
+
+
 def _reconcile_fcst_proportions(
     S: np.ndarray,
     y_hat: np.ndarray,
@@ -399,6 +487,13 @@ def _reconcile_fcst_proportions(
     nodes: dict[str, dict[int, np.ndarray]],
     idxs_top: np.ndarray,
 ):
+    """Pure-Python reference for the forecast-proportions traversal.
+
+    The runtime path is `_reconcile_fcst_proportions_native`; this
+    implementation is kept as the behavioral reference the regression tests
+    assert bit-identical equality against, and as the runtime fallback for
+    non-float64 input (the native kernel is float64-only).
+    """
     reconciled = np.zeros_like(y_hat)
     level_names = list(tags.keys())
     for idx_top in idxs_top:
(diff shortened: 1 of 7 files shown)


loama:

